PR #114, opened 2017-07-06
AMLS-2093: count partners in responsible people and redirect if Partn…

Froom2:
…ership + basic get method


conf/app.routes
```diff
@@ -273,6 +273,8 @@ GET         /update-information/remove-responsible-person
 
 GET         /declaration/register-responsible-person                                  @controllers.declaration.RegisterResponsiblePersonController.get()
 
+GET         /update-information/partner                                               @controllers.declaration.RegisterPartnersController.get()
```
elkdanger: In keeping with the other routes in this flow, could we change this to be `/declaration/partner`?

PR #141, opened 2017-07-18
Amls 2093 - fixed build failures

lmugnai:


test/services/ProgressServiceSpec.scala
```diff
@@ -29,15 +29,28 @@ import org.mockito.Mockito._
...
 import uk.gov.hmrc.play.http.HeaderCarrier
+import utils.GenericTestHelper
 
 import scala.concurrent.ExecutionContext.Implicits._
 import scala.concurrent.{ExecutionContext, Future}
 
-class ProgressServiceSpec extends PlaySpec with MockitoSugar with ScalaFutures with OneAppPerSuite {
+class ProgressServiceSpec extends GenericTestHelper with MockitoSugar with ScalaFutures with OneAppPerSuite {
+
+  val paymentsConnector = mock[PayApiConnector]
```
elkdanger: Would it be possible to take this out? ProgressService won't use the payments connector and this could confuse things with it being in the test.

test/services/ProgressServiceSpec.scala
```diff
@@ -29,15 +29,28 @@ import org.mockito.Mockito._
...
+
+  implicit override lazy val app: Application = new GuiceApplicationBuilder()
+    .disable[com.kenshoo.play.metrics.PlayModule]
+    .bindings(bindModules: _*).in(Mode.Test)
+    .bindings(bind[PayApiConnector].to(paymentsConnector))
```
elkdanger: This will also need to come out as a result of removing line 46

test/services/ProgressServiceSpec.scala
lmugnai: done

test/services/ProgressServiceSpec.scala
lmugnai: removed

PR #331, opened 2017-10-23
Amls 3119

darrylmurphy:


app/utils/ControllerHelper.scala
```diff
@@ -110,8 +114,16 @@ object ControllerHelper {
...
   def hasNominatedOfficer(eventualMaybePeoples: Future[Option[Seq[ResponsiblePeople]]]): Future[Boolean] = {
     eventualMaybePeoples map {
-      case Some(rps) =>  rps.filter(!_.status.contains(StatusConstants.Deleted)).exists(_.positions.fold(false)(_.positions.contains(NominatedOfficer)))
-      case _ =>  false
+      case Some(rps) =>  rps.filter(!_.status.contains(StatusConstants.Deleted)).exists(_.isNominatedOfficer)
+      case _ => false
+    }
+  }
+
+  def getNominatedOfficer(responsiblePeople: Future[Option[Seq[ResponsiblePeople]]]): Future[Option[ResponsiblePeople]] = {
```
elkdanger: Would it be possible to rework the signature on this to just take `Seq[ResponsiblePeople]` instead of `Future[Option[Seq[ResponsiblePeople]]]`?

Reason being that it would make the function much more versatile, were we to use it again later where we might not be in the context of a `Future[Option[T]]`. Means that whereever we have a simple `Seq[ResponsiblePeople]` we would always be able to find the nominated officer without having to wrap it.

I appreciate it would have an impact on the call in `StatusController` but I think it would stand us in good stead going forward.

PR #398, opened 2017-11-23
AMLS-3127: Remove corporation tax from check your answers in about the business

darrylmurphy:


elkdanger:
Does this relate to AMLS-3127? If so, shouldn't this question only be hidden on 'Check your answers' only if it hasn't been auto-matched in Business Matching?

darrylmurphy:
The UTR has been removed for all cases as we believe a user would not be matched without their UTR - this means the UTR will always be matched when going through our application, therefore never be asked and never be shown in Check Your Answers.

PR #401, opened 2017-11-24
AMLS-3015: Page split: Do you keep records of all your transactions with customers

elkdanger:
* Backwards compatible JSON readers
* Fixed routing for yes/no edit modes
* Changed form reader for yes/no boolean type
* Refactored original yes/no page to support DI
* Tests for backwards compatibility
* Removed old page model
* Created generic `constant[A]` Json reads helper in `utils.MappingUtils`


test/models/businessactivities/BusinessActivitiesSpec.scala
```diff
@@ -98,11 +101,46 @@ class BusinessActivitiesSpec extends PlaySpec with MockitoSugar with OneAppPerSu
...
   )
 
   val completeJson = Json.obj(
+    "involvedInOther" -> true,
+    "details" -> DefaultInvolvedInOtherDetails,
+    "expectedBusinessTurnover" -> "01",
+    "expectedAMLSTurnover" -> "01",
+    "businessFranchise" -> true,
+    "franchiseName" -> DefaultFranchiseName,
+    "isRecorded" -> true,
```
lmugnai: Should we have a test for when this is None? Or are we confident this is always going to be set as true or false?

app/models/businessactivities/TransactionType.scala
```diff
@@ -65,6 +66,30 @@ object TransactionTypes {
...
+          case "03" =>
+            (__ \ "digitalSoftwareName").read[String].map (DigitalSoftware.apply  _) map identity[TransactionType]
+          case _ =>
+            Reads(_ => JsError((__ \ "transactions") -> play.api.data.validation.ValidationError("error.invalid")))
+        }.foldLeft[Reads[Set[TransactionType]]](
+          Reads[Set[TransactionType]](_ => JsSuccess(Set.empty))
+        ){
+          (result, data) =>
+            data flatMap {m =>
+              result.map {n =>
```
lmugnai: Did you decide against re-writing this in a more readable way? :)

elkdanger: I think when we were chatting about I was referring to the JSON readers for the [equivalent type in the backend microservice](https://github.com/hmrc/amls/blob/master/app/models/fe/businessactivities/TransactionRecord.scala) but this one can definitely be rewritten in a similar fashion. Would you mind if I put it in a separate pull request so that this can be tested in the meantime?

test/models/businessactivities/BusinessActivitiesSpec.scala
elkdanger: Agreed, this should be tested as `None` is definitely a possibility; I'll push up a new commit which does so.

app/models/businessactivities/TransactionType.scala
lmugnai: Yep.

test/models/businessactivities/BusinessActivitiesSpec.scala
elkdanger: Looking at the spec more closely, I think the `None` case will already be covered by the tests that validate a partial Json packet

https://github.com/hmrc/amls-frontend/blob/eb9ef660ac6d0a7ba89bc3d80871cf7706588ba7/test/models/businessactivities/BusinessActivitiesSpec.scala#L211

PR #419, opened 2017-12-05
Share filter functionality to remove deleted or empty rp from seq, re…

darrylmurphy:
…move hasAccepted toggle from BankDetails

### DIFF
--- a/app/models/responsiblepeople/ResponsiblePeople.scala
+++ b/app/models/responsiblepeople/ResponsiblePeople.scala
@@ -190,12 +190,12 @@ object ResponsiblePeople {
 
   implicit val formatOption = Reads.optionWithNull[Seq[ResponsiblePeople]]
 
+  def filter(rp: Seq[ResponsiblePeople]) = rp.filterNot(_.status.contains(StatusConstants.Deleted)).filterNot(_ == ResponsiblePeople())
+
   def section(implicit cache: CacheMap): Section = {
 
     val messageKey = "responsiblepeople"
     val notStarted = Section(messageKey, NotStarted, false, controllers.responsiblepeople.routes.ResponsiblePeopleAddController.get())
-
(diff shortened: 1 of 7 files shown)


darrylmurphy:
AMLS-3054

Bug created by lack of filter of deleted or empty. Solution will be to use this shared filter.

app/models/responsiblepeople/ResponsiblePeople.scala
```diff
@@ -190,13 +190,13 @@ object ResponsiblePeople {
 
   implicit val formatOption = Reads.optionWithNull[Seq[ResponsiblePeople]]
 
+  def filter(rp: Seq[ResponsiblePeople]) = rp.filterNot(_.status.contains(StatusConstants.Deleted)).filterNot(_ == ResponsiblePeople())
```
elkdanger: Now that the scope of this method has changed and is more generally available, could we rename it? Maybe something like `filterDeleted`, or `cleanup`?

I'm just conscious that `filter` doesn't really tell enough about what it filters. Do you have any strong opinions about it?

darrylmurphy: Perhaps? Although it'll filter deleted and empty RPs at the moment, does filterDeleted stray from the truth and cleanup == filter?
Does it sound fair enough that "filter" used in a general context imply it'll remove RPs such as deleted and other unwanted criteria?

elkdanger: I suppose as long as we're generally aware of what it does. Maybe if confusion arises from it in the future, we can look at refactoring it a bit to make it more clear.

darrylmurphy: I agree, I think its trying to balance in the name between being descriptive but flexible. I certainly don't feel strongly about it, so I'm open to change if it proves unsuitable.

PR #464, opened 2018-01-02
AMLS-3266 Remove Toggles - change-officer

lmugnai:


darrylmurphy:
I think it'll be worth removing ChangeOfficerFeatureFilter (https://github.com/hmrc/amls-frontend/blob/48e41c6da861f0f7cd0f5ec5b4e357ae8042e708/app/filters/ChangeOfficerFeatureFilter.scala) which is there specifically for the change-officer toggle. 
Also the references to it in the spec?

PR #580, opened 2018-05-11
Amls 3630

zatevad:


app/controllers/businessmatching/updateservice/RemoveServiceHelper.scala
```diff
@@ -0,0 +1,112 @@
...
+import scala.concurrent.ExecutionContext.Implicits.global
+
+@Singleton
+class RemoveServiceHelper @Inject()(val authConnector: AuthConnector,
+                                    implicit val dataCacheConnector: DataCacheConnector
+                                   ) {
+  def removeBusinessActivities(model: RemoveServiceFlowModel)(implicit ac: AuthContext, hc: HeaderCarrier, ec: ExecutionContext): Future[Unit] = {
+    ???
+  }
+
```
elkdanger: Could we remove this if it's no longer going to be needed?

app/models/flowmanagement/RemoveServiceFlowModel.scala
```diff
@@ -16,6 +16,15 @@
 
 package models.flowmanagement
 
-case class RemoveServiceFlowModel(dummy: Int)
+import models.businessmatching.BusinessActivity
+import play.api.libs.json.Json
 
-object RemoveServiceFlowModel
+case class RemoveServiceFlowModel(existingActivities: Option[Set[BusinessActivity]] = None,
+                                  activitiesToRemove: Option[Set[BusinessActivity]] = None)
```
elkdanger: Is `existingActivities` being used, and could we not simply retrieve this information from the business matching data?

PR #587, opened 2018-05-15
Amls 3474

kerryannekane:


app/controllers/LandingController.scala
```diff
@@ -213,19 +212,18 @@ class LandingController @Inject()(val landingService: LandingService,
...
+          fixEmpties flatMap { cacheMap =>
             if (dataHasChanged(cacheMap)) {
-              (cacheMap.getEntry[SubscriptionResponse](SubscriptionResponse.key),
-                cacheMap.getEntry[AmendVariationRenewalResponse](AmendVariationRenewalResponse.key)) match {
-                case (Some(_), _) => refreshAndRedirect(amlsRegistrationNumber, Some(cacheMap))
-                case (_, Some(_)) => refreshAndRedirect(amlsRegistrationNumber, Some(cacheMap))
-                case _ => setAlCorrespondenceAddressAndRedirect(amlsRegistrationNumber, Some(cacheMap))
-
+              cacheMap.getEntry[SubmissionRequestStatus](SubmissionRequestStatus.key) collect {
+                case SubmissionRequestStatus(true) => refreshAndRedirect(amlsRegistrationNumber, Some(cacheMap))
```
JGRobinson: Have we lost the AmendVariationRenewalResponse????

elkdanger: The original logic here relied on either `SubscriptionResponse` or `AmendVariationRenewalResponse` data being in the cache in order to be able to tell us whether or not the user has just done a submission. Originally, this data was not retained when the cache was cleared and so this logic is fine.

Now though, this data _is_ being retained when the cache is cleared and so this logic here is no longer a reliable indicator as to whether the user has just submitted an application or not. Thus, that has been replaced by a new type `SubmissionRequestStatus` which is trashed when the data cache is cleared immediately following a submission.

Happy to discuss.

elkdanger: Happy to merge. There's a tiny thing we can do in the messages file concerning passing the link into it but it's very minor and I'll pick that up with you later.

PR #599, opened 2018-05-17
Amls 3783

zatevad:


elkdanger:
The fundamentals of this PR rely on the fact that it hides the 'remove a business type' radio button from the user when there is only 1 business type available to be removed, but that doesn't follow the pre-agreed flow, which is:

* The user (regardless of the amount of business types) is allowed to select the 'remove' option
* If they only have one business type to remove (violating business logic) then a content page is shown to them which explains why their one and only business type cannot be removed from the application

Given that, would it be possible to refactor out the `allowRemove` predicates and allow the user to see the radio button at all times? The flow routing that has been written should already take care of the rest.

PR #630, opened 2018-05-25
Amls 3500 trading premises playing back wrong address

JGRobinson:


app/controllers/tradingpremises/IsResidentialController.scala
```diff
@@ -57,7 +57,7 @@ class  IsResidentialController @Inject()(
               address <- getAddress(bm) if isFirstTradingPremises(tradingPremises, index)
             } yield address
 
-            Ok(is_residential(form, address, index, edit))
+            Ok(is_residential(form, Some(tp.yourTradingPremises.get.tradingPremisesAddress.toBCAddress), index, edit))
```
elkdanger: The second parameter here could be rewritten using a map, which would get rid of the Some() part and also the `.get` call, which is bad practise.

For example: `Ok(is_residential(form, tp.yourTradingPremises.map(_.tradingPremisesAddress.toBCAddress)...`

This would prevent exceptions occurring in the event that `yourTradingPremises` happens to be `None`, which is catered for earlier in the code.

app/controllers/tradingpremises/IsResidentialController.scala
```diff
@@ -57,7 +57,7 @@ class  IsResidentialController @Inject()(
               address <- getAddress(bm) if isFirstTradingPremises(tradingPremises, index)
             } yield address
```
elkdanger: This constructing of the `address` val could be removed completely as it's no longer used.

PR #639, opened 2018-06-01
AMLS-3637: Implementation of dynamic replaceable Save4Later data

zatevad:


app/services/UpdateSave4LaterService.scala
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+
+  def fn[T](key: String, m: Option[T])(implicit ac: AuthContext, hc: HeaderCarrier, fmt: Format[T]): Future[CacheMap] = m match {
+    case Some(model) => cacheConnector.save[T](key, model)
+    case _ => Future.successful(CacheMap("", Map.empty))
+  }
+
+  def getSaveForLaterData(fileName: String)(implicit hc: HeaderCarrier, ac: AuthContext, ex: ExecutionContext): Future[Option[UpdateSave4LaterResponse]] = {
+    val requestUrl = s"http://localhost:8941/anti-money-laundering/saveforlater/getfile/${fileName}"
```
elkdanger: Could this absolute url be pulled out and be derived from configuration, like the other services that we use? We'll need to add a new config entry for stubs as there isn't one currently.

test/services/UpdateSave4LaterServiceSpec.scala
```diff
@@ -0,0 +1,264 @@
...
+        Mockito.verify(mockCacheConnector).save[Supervision](eqTo(Supervision.key), any())(any(), any(), any())
+        Mockito.verify(mockCacheConnector).save[AboutTheBusiness](eqTo(AboutTheBusiness.key), any())(any(), any(), any())
+        Mockito.verify(mockCacheConnector).save[EstateAgentBusiness](eqTo(EstateAgentBusiness.key), any())(any(), any(), any())
+        Mockito.verify(mockCacheConnector).save[SubscriptionResponse](eqTo(SubscriptionResponse.key), any())(any(), any(), any())
+        Mockito.verify(mockCacheConnector).save[AmendVariationRenewalResponse](eqTo(AmendVariationRenewalResponse.key), any())(any(), any(), any())
+      }
+    }
+
+    "getDataFromStubs is called" must {
+      "return the specified file contents from stubs" in new Fixture {
```
elkdanger: There is a missing test here. If the test is no longer needed, it should be removed from the code.

PR #663, opened 2018-06-08
AMLS-3552 | Removed 4th button from bankaccounttypes view

matthewharris1:


app/views/bankdetails/bank_account_types.scala.html
```diff
@@ -55,13 +55,6 @@
...
                     value = "03"
                 )
-                @if(count <= 1) {
-                    @radio(
-                        f = f("bankAccountType"),
-                        labelText = "bankdetails.accounttype.lbl.04",
-                        value = "04"
-                    )
-                }
             }
```
JGRobinson: There are additional tests in here which relate to if a person already has bank accounts which need to be removed.

test/controllers/bankdetails/BankAccountTypeControllerSpec.scala
```diff
@@ -74,8 +74,7 @@ class BankAccountTypeControllerSpec extends AmlsSpec with MockitoSugar {
           document.select("input[type=radio][name=bankAccountType][value=01]").hasAttr("checked") must be(false)
           document.select("input[type=radio][name=bankAccountType][value=02]").hasAttr("checked") must be(false)
           document.select("input[type=radio][name=bankAccountType][value=03]").hasAttr("checked") must be(false)
-          document.select("input[type=radio][name=bankAccountType][value=04]").hasAttr("checked") must be(false)
-          document.select("input[type=radio][name=bankAccountType]").size() must be(4)
+          document.select("input[type=radio][name=bankAccountType]").size() must be(3)
```
JGRobinson: There are additional tests in this file in relation to when items exist which need to be removed.

PR #664, opened 2018-06-08
AMLS-3893 Check your answers page

nonimay:
**WARNING**: There is still work do be done around the entering a section. This is because a section redirects to the old CYA page which was more of an index page. CYA now requires an index and so the section code takes in -1. This currently will stop users from being able to get into the bank accounts section if they have more than one account complete/incomplete.

CYA page now for one bank account only


conf/messages
```diff
@@ -925,7 +926,7 @@ aboutthebusiness.registeredforcorporationtax.taxReference = Corporation Tax refe
 aboutthebusiness.registeredforcorporationtax.taxReferenceHint = It’s 10 numbers with no letters, punctuation or other characters.
 aboutthebusiness.contactingyou.email.title = What’s your business’s email address?
 aboutthebusiness.contactingyou.phone.title = What’s your business’s phone number?
-aboutthebusiness.contactingyou.lbl = We’ll email you when there’s an update about your application for anti-money laundering supervision.
+aboutthebusiness.contactingyou.lbl = We’ll email you when there’s an updafaccate about your application for anti-money laundering supervision.
```
elkdanger: Spelling error here

app/controllers/bankdetails/SummaryController.scala
```diff
@@ -16,63 +16,52 @@
...
+                                 ) extends BaseController with RepeatingSection {
 
-  private def updateBankDetails(bankDetails: Option[Seq[BankDetails]]) : Future[Option[Seq[BankDetails]]] = {
+  private def updateBankDetails(bankDetails: Option[Seq[BankDetails]], index: Int) : Future[Option[Seq[BankDetails]]] = {
     bankDetails match {
       case Some(bdSeq) => {
-        val updatedList = bdSeq.map { bank =>
-          bank.copy(hasAccepted = true)
+        val updatedList = bdSeq.zipWithIndex.map {
+          case (bank, i) => if (i == index -1) bank.copy(hasAccepted = true) else bank
```
elkdanger: We can use `RepeatingSection.updateData` here to update this instead of working out the index ourselves.

PR #665, opened 2018-06-08
Amls 3552 remove register another account page

matthewharris1:
Was unsure as to which page to change the redirect to but I assumed it would be the new YourBankAccountsController. Let me know if I'm wrong!


app/controllers/bankdetails/BankAccountIsUKController.scala
```diff
@@ -81,7 +81,7 @@ class BankAccountIsUKController @Inject()(
               if (edit) {
                 Future.successful(Redirect(routes.SummaryController.get(false)))
               } else {
-                lazy val redirect = Redirect(routes.BankAccountRegisteredController.get(index))
+                lazy val redirect = Redirect(routes.YourBankAccountsController.get())
```
JGRobinson: Can you have a chat to Fi about this - it should be going to check your answers and have an index associated with it - but she may already have changed it.

elkdanger: @nonimay ^^

PR #668, opened 2018-06-11
Amls 3892 your bank accounts page

JGRobinson:


app/views/bankdetails/your_bank_accounts.scala.html
```diff
@@ -33,65 +33,90 @@
...
-        <h3>@Messages("bankdetails.yourbankaccounts.incomplete")</h3>
+        <h3 id="incomplete-header">@Messages("bankdetails.yourbankaccounts.incomplete")</h3>
         <ul class="list">
         @inCompleteModel.map { bankDetail =>
-            <li class="group grid-layout">
-                <div class="grid-layout__column grid-layout__column--3-4">
-                    <div class="grid-layout__column grid-layout--2-3">
+            <li class="group summary__listing">
+                <div>
+                    <div id="incomplete-detail-@bankDetail._2" class="grid-layout__column grid-layout--2-3">
```
elkdanger: Could we use some case-matching here to remove the `_1`'s and `_2`'s?

PR #671, opened 2018-06-11
Amls 3907 update bank account name page

matthewharris1:


app/controllers/bankdetails/BankAccountNameController.scala
```diff
@@ -39,46 +39,57 @@ class BankAccountNameController @Inject()(
...
-              if (edit) {
-                Redirect(routes.SummaryController.get(index))
-              } else {
-                Redirect(routes.BankAccountIsUKController.get(index))
+                  Redirect(routes.BankAccountTypeController.get(i))
+                }
+              }
+              case _ => dataCacheConnector.fetch[Seq[BankDetails]](BankDetails.key) flatMap { maybeBankDetails =>
+                val newList = maybeBankDetails.getOrElse(Seq.empty) ++ Seq(newBankDetails)
+                dataCacheConnector.save(BankDetails.key, newList) map { _ => Redirect(routes.BankAccountTypeController.get(index.get)) }
```
elkdanger: An issue I've just come across is that the index value that is passed to the `BankAccountTypeController.get` must be the index of the newly created instance of `BankDetails`. In this case, the index you want is the same value as `newList.size`, as it's just the index of the last inserted index.

elkdanger:
There is a further issue with this entire controller that it needs to be split into indexed + non-indexed routes, as just having an `Option[Int]` on the parameter list isn't enough - Play will return a 404.

See AMLS-3913 for details.

PR #680, opened 2018-06-12
AMLS-3928 CYA page goes to your bank accounts page on submit

nonimay:


app/views/bankdetails/summary.scala.html
```diff
@@ -100,6 +101,10 @@
         }
     }
 
+    @form(EmptyForm, controllers.bankdetails.routes.SummaryController.post(index)) {
+        @submit(false, Some(Messages("button.checkyouranswers.acceptandaddbankaccount")), false)
+    }
+
     @anchor(
```
elkdanger: Should this anchor come out? Will it not leave a second button on the page that looks very similar to the form button?

nonimay: yes...

PR #720, opened 2018-06-27
AMLS-4062 | BUG - Renewal amendment is showing an accumulative amount

drewmiley:


drewmiley:
AMLS-4063 | BUG - Renewal amendments - no fee shows a fee
should also be fixed by this

PR #740, opened 2018-07-09
AMLS-4079 Migrated Save4Later to Mongo

elkdanger:
This PR implements everything to get the cache working out of Mongo.

**Note:**
* The feature is currently **turned off** (see `application.conf:appCache.mongo.enabled`)
* There is currently an issue with Mongo connections being dropped whenever code is changed, forcing the developer to restart their app process. This only manifests if the feature is switched **on**


elkdanger:
@ajchilts @JGRobinson Let's go through this in detail tomorrow.

PR #753, opened 2018-07-17
AMLS-4079 Altered cache migration strategy

elkdanger:
Previously, data was migrated when loaded and the data was primarily loaded from the mongo cache. This commit changes that so that the data is primarily loaded from S4L, and the data is migrated slowly on save.

At some point later (after 28 days has passed), we're going to have to update this so that the mongo store becomes the primary cache and the data is dropped from S4L.

This has been done in attempt to fix multiple issues with migration, especially when adding new services through the PAV flow.


elkdanger:
Don't merge this yet please - not sure it's the right approach.

elkdanger:
@ajchilts This is ready for review and merge now

PR #769, opened 2018-08-01
AMLS-4096 | Added FX to CYA page - Refactored Unit tests

matthewharris1:
# PR Details

Added FX to the CYA renewals page

Added FX to the summary view and refactored unit tests to include this

- [x] Build passing.

### DIFF
--- a/test/views/renewal/summarySpec.scala
+++ b/test/views/renewal/summarySpec.scala
@@ -98,7 +98,7 @@ class summarySpec extends AmlsSpec
           Some(SendTheLargestAmountsOfMoney(Country("France", "FR"))),
           Some(MostTransactions(Seq(Country("United Kingdom", "GB")))),
           Some(CETransactionsInLast12Months("123")),
-          None,
+          Some(FXTransactionsInLast12Months("12")),
           false
         )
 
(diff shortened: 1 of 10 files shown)


test/views/renewal/summarySpec.scala
```diff
@@ -98,7 +98,7 @@ class summarySpec extends AmlsSpec
           Some(SendTheLargestAmountsOfMoney(Country("France", "FR"))),
           Some(MostTransactions(Seq(Country("United Kingdom", "GB")))),
           Some(CETransactionsInLast12Months("123")),
-          None,
+          Some(FXTransactionsInLast12Months("12")),
```
elkdanger: It's good that we've added an FX value to the test data for the view spec, but are we asserting that the correct value is written out on the page?

Further down the spec there is a 'section checks' block which actually checks that the questions are being written out on the page with the correct headings and values - could we add something in there to check that the FX question is being written out?

Have a look at these links to see what I mean:
https://github.com/hmrc/amls-frontend/blob/974c511d0d0b4412321b5720e2ea1b9b9f94f0ee/test/views/renewal/summarySpec.scala#L71
https://github.com/hmrc/amls-frontend/blob/974c511d0d0b4412321b5720e2ea1b9b9f94f0ee/test/views/renewal/summarySpec.scala#L129

matthewharris1: Added the section check for FX but also one for CE.

PR #803, opened 2018-09-19
Missing tests for existing breakdown row table functionality

drewmiley:
## Checklist

- [x] Build passing.
- [x] Unit test approach and implementation has been reviewed with QA (testers).

### DIFF
--- a/app/views/confirmation/confirmation_new.scala.html
+++ b/app/views/confirmation/confirmation_new.scala.html
@@ -18,6 +18,7 @@
 @import forms2._
 @import models.confirmation._
 @import config.ApplicationConfig
+@import views.html.confirmation.{breakdown => breakdown_display}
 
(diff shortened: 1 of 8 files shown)


app/views/confirmation/confirmation_new.scala.html
```diff
@@ -18,6 +18,7 @@
 @import forms2._
```
JGRobinson: Do we need to do this refactor now ?

JGRobinson: Do we need to do this refactor now.

drewmiley:
@JGRobinson 👍 This should form part of the work for AMLS-3402/AMLS-4218 due to refactor of the breakdown display.

ajchilts:
I am going to close this PR as the code base has moved on a great deal since it was raised. We will however; keep the branch should we need to come back to this. 

As this is the fee breakdown table only I am reluctant to spend too much time on this. We can address test coverage of the table at some point in the future when we have less pressing issues.

PR #882, opened 2019-01-02
Amls 3564

owen-opencast:
Implemented a cache-busting strategy which will cause browsers to re-download assets when they have been changed.

Used standard Play Framework functionality to implement Asset Fingerprinting. Asset fingerprinting adds an ID number to asset filenames based on the contents of the file, so that if files are changed the filename is also updated, preventing cacheing.

Solution was based on previous HMRC use of Play Asset Fingerprinting [here](https://github.com/hmrc/cgt-calculator-non-resident-frontend/pull/124/files).

## Checklist

- [x] Build passing.


ajchilts:
Passing automation. 
screen shot 2019-01-07 at 11 41 17

PR #925, opened 2019-02-08
Amls 4637 businessactivities controller DI

samzaroughmrc:

## Checklist

- [ ] Build passing.

### DIFF
--- a/app/controllers/businessactivities/AccountantForAMLSRegulationsController.scala
+++ b/app/controllers/businessactivities/AccountantForAMLSRegulationsController.scala
@@ -16,19 +16,20 @@
 
 package controllers.businessactivities
 
-import config.AMLSAuthConnector
 import connectors.DataCacheConnector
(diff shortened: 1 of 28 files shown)


owen-opencast:
Just talked with @ajchilts - recommendation is that we don't start adding singleton annotations until performance requires them. Can we remove them?

slavvco:
Isn't it actually good approach to use @Singleton annotation? https://www.playframework.com/documentation/2.6.x/ScalaDependencyInjection#Singletons

owen-opencast:
The documentation from that link recommends the use of Singletons when you have state, or when you're caching things within the class - I don't think the controllers should be caching anything, they should just be doing simple processing and forwarding of requests.

There may be a slight performance benefit from reusing a single controller object, but current recommendation is we don't use them until we need to.

owen-opencast:
Passes spec and regression tests, merging into master.

PR #987, opened 2019-03-27
AMLS-4768 | Updated TCSP Content issues

matthewharris1:
TCSP content changes

## How Has This Been Tested?
Manual Testing
Unit Testing

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


conf/messages
```diff
@@ -1762,37 +1762,38 @@ responsiblepeople.country.of.birth.heading = Was {0} born in the UK?
...
-tcsp.provided_services.service.lbl.05 = Mail forwarding
-tcsp.provided_services.service.lbl.06 = Receptionist
-tcsp.provided_services.service.lbl.07 = Conference rooms
+tcsp.provided_services.service.lbl.04 = Mail forwarding
+tcsp.provided_services.service.lbl.05 = Self-collect mailboxes
+tcsp.provided_services.service.lbl.06 = Telephone call handling
+tcsp.provided_services.service.lbl.07 = Receptionist
 tcsp.provided_services.service.lbl.08 = Other
-tcsp.provided_services.service.lbl.details = Please give details
+tcsp.provided_services.service.lbl.details = Service type
```
slavvco: Above you have changed the labels... so now user will select for example 'Conference room' but this will be translated to 'Phone handling' anyway because model is constructed like that... so user submitting application with 'Conference room' will eventually get back the 'Phone handling'. So it cannot be done like that I think...

matthewharris1: Sorted :+1:

test/views/tcsp/summarySpec.scala
```diff
@@ -70,7 +70,7 @@ class summarySpec extends AmlsSpec with MustMatchers with HtmlAssertions with Ta
         val testdata = Tcsp(
           Some(TcspTypes(Set(NomineeShareholdersProvider, TrusteeProvider, RegisteredOfficeEtc, CompanyDirectorEtc, CompanyFormationAgent(true,false)))),
           Some(ProvidedServices(Set(
-            PhonecallHandling,EmailHandling,EmailServer,SelfCollectMailboxes,MailForwarding,Receptionist,ConferenceRooms, Other("sfasfasef")
+            PhonecallHandling,EmailHandling,EmailServer,SelfCollectMailboxes,MailForwarding,Receptionist,ConferenceRooms, Other(" sfasfasef")
```
slavvco: Everything looks good, I'm just not convinced that in above line should be the space in front of sfasfasef...

app/views/tcsp/summary.scala.html
```diff
@@ -87,8 +98,17 @@
             ) {
                 <ul class="list list--comma">
                     @model.providedServices.map { a =>
-                        @a.services.map { x =>
-                            <li>@x.getMessage</li>
+                        @if(a.services.collectFirst {case Other(_) => true}){
```
drewmiley: Do we need this if/else? I think that just having lines 102-107 might achieve the same outcome?

app/views/tcsp/summary.scala.html
```diff
@@ -38,14 +39,24 @@
...
         <ul class="list list--semicolon">
-            @v.serviceProviders.map { x =>
-                <li>@x match {
-                    case NomineeShareholdersProvider => {@Messages("tcsp.service.provider.lbl.01")}
-                    case TrusteeProvider => {@Messages("tcsp.service.provider.lbl.02")}
-                    case RegisteredOfficeEtc => {@Messages("tcsp.service.provider.lbl.03")}
-                    case CompanyDirectorEtc => {@Messages("tcsp.service.provider.lbl.04")}
-                    case CompanyFormationAgent(_, _) => {@Messages("tcsp.service.provider.lbl.05")}
-                }</li>
+            @if(v.serviceProviders.contains(CompanyDirectorEtc)) {
```
drewmiley: There is probably a cleverer way to do this, but definitely not as clear. I like it 👍

samzaroughmrc: use list.sortedBy please!

app/controllers/tcsp/SummaryController.scala
```diff
@@ -37,11 +38,28 @@ class SummaryController @Inject()
...
+
+    val special = (data.tcspTypes match {
+      case Some(types) => types.serviceProviders.filter { case provider => provider.value == "05" }
+    }) map {
+      case _ => Messages("tcsp.service.provider.lbl.05")
+    }
+
+    res.toList.sorted ++ special.toList
+  }
+
```
slavvco: what if tcspTypes is empty? data.tcspTypes matching will fail. Will you pass then empty list? Will the view display the question with empty answer?

app/views/tcsp/summary.scala.html
slavvco: this was to ensure that other will appear as last if there are some other provided services, now I'm not sure what was in lines 102-107 as everything moved but if you meant that was was in the body of if part only, I think you are right. This may be sorted as well in controller as @samzaroughmrc suggested

app/controllers/tcsp/SummaryController.scala
samzaroughmrc: this should never happen as the frontend wont allow the user to select nothing.
we can be defensive in out coding, but if this turn out to cause a problem, then the issue is with the ui page validators not the responsibility of this function

app/views/tcsp/summary.scala.html
slavvco: done

app/controllers/tcsp/SummaryController.scala
slavvco: at the moment it is possible to get into situation where summary page will try to display the questions and answers even when there's no answers in database, it's possible using back button:
1. add tcsp and answer all questions
2. remove tcsp
3. go back using back button to check your answers
you will see questions with empty answers

this is problem with check your answers itself which should not be displayed if section is incomplete, but proves that our frontend is not bulletproof

slavvco: so we could leave it as it is, because we have ticket in backlog to address this issue, but as long as it's not solved I don't think we should leave this option to display empty answers

samzaroughmrc: the back-link problem does not happen with latest code merged from master

samzaroughmrc: I understand the effect this can have if the list if empty.
coming from the mindset of **single responsibility principle**, I would say something else should handle the fact that the list is empty to start with. hence, the sortedList function should not get called. 
However, in this case we have few protection layers,

- the frontend wont allow you to leave the page without selecting at least one.
- amls wont allow you submit as the section will be incomplete.

hope that answers your concerns

slavvco: sortedList function above can return List of strings or empty list... interpreting the signature of the function it should return sorted list... what is then the default order of empty list ascending or descending? if you cannot answer it means this function is doing two things: returning sorted list or empty list, so there are two possible results, so single responsibility is violated here as well :) should this be handled somehow before calling the sortedList function... maybe but it's not. Unfortunately model is constructed so that the TcspTypes are optional, so empty list is possible and valid outcome and the schema accepts that happily...

samzaroughmrc: @ajchilts ^^

app/controllers/tcsp/SummaryController.scala
```diff
@@ -37,11 +38,29 @@ class SummaryController @Inject()
...
+    val sortedList = for {
+      types <- data.tcspTypes
+      providers <- Some(types.serviceProviders)
+      labels <- Some(providers.collect {
+          case provider if provider.value != "05" => Messages(s"tcsp.service.provider.lbl.${provider.value}")
+        }
+      )
+      specialCase <- Some(providers.collect {
+          case provider if provider.value == "05" => Messages(s"tcsp.service.provider.lbl.05")
```
owen-opencast: Optional readability suggestion - could we use `provider.isInstanceOf[CompanyFormationAgent]` instead of `provider.value == "05"`? (Normally would avoid isInstanceOf, but ServiceProvider is a sealed trait...)

app/controllers/tcsp/SummaryController.scala
ajchilts: Ultimately test the solution. If we identify a scenario that breaks the code we should either; 

i). Fix the scenario so it is not possible to break the code or;
ii). Add some defensive coding to handle it. 
iii). Both!

Despite what is stated in various standards I have seen too many instances where we get caught out by a scenario that 'should not exist'. With this in mind I do tend to code defensively. Remember; the front end is not the only manipulator of data! We have ETMP in play and we have been stung by schema mismatches and missing mongo data. I do not have a problem with devs coding defensively as 500 'please try again later' messages are a bad user experience. 

While standards are here for a reason and very important please put the user journey at the forefront of you mind when coding solutions. Of course also adhering to the standards where possible.

app/controllers/tcsp/SummaryController.scala
samzaroughmrc: good point @owen-opencast

samzaroughmrc: refactored

app/controllers/tcsp/SummaryController.scala
matthewharris1: Following andys comment we should then revert to swavs commit and pass this ticket on since it covers all cases? In my opinion to much time has been spent on this ticket already. The functionality worked in the first place, passed tests and was secure. We've spent an extra day working on this.

samzaroughmrc: a quick mobbing session with everyone, we decided to keep the unit tested code and log an error if the list is empty

drewmiley:
I considered all this in my initial review.

owen-opencast:
Moving to testing

PR #994, opened 2019-03-29
AMLS-4823 | Updated css to not display double line for newly added sections to registration progress

slavvco:
As in the subject. Updated frontend css to not display double line when adding new service to application.

## Related / Dependant PRs?

none

## How Has This Been Tested?

manually

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.

### DIFF
--- a/app/views/registrationamendment/section_list.scala.html
+++ b/app/views/registrationamendment/section_list.scala.html
@@ -19,19 +19,13 @@
 @import views.html.registrationamendment.section
 
 @(sections: Seq[Section], newSections: Option[Seq[Section]] = None)(implicit request: Request[_], m: Messages)
-
     @* new sections *@
-    @newSections.collect { case seq if seq.nonEmpty =>
-        <ul id="new-sections-list" class="grid-layout__column grid-layout__column--2-3 display-block">
-            @seq.map { s =>
-                @submissionSection(s.name, s.status, s.call)
-            }
-        </ul>
-    }
+
+    @allSections = @{sections union newSections.getOrElse(Seq())}
 
     @* existing *@
     <ul id="existing-sections-list" class="grid-layout__column grid-layout__column--2-3 display-block">
-        @sections.map { case Section(name, status, hasChanged, call) =>
-            @section(name, status, hasChanged, call)
+        @allSections.sortBy(section => Messages(s"progress.${section.name}.name")).map { s =>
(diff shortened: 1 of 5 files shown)


slavvco:
going back to in progress as the new sections shouldn't appear above the old ones anymore

owen-opencast:
Ready for review again - I've merged the two sections lists, keeping the existing text (Add / Manage etc). Passes spec and acceptance tests.

app/views/registrationamendment/section_list.scala.html
```diff
@@ -19,19 +19,13 @@
...
+
+    @allSections = @{sections union newSections.getOrElse(Seq())}
 
     @* existing *@
     <ul id="existing-sections-list" class="grid-layout__column grid-layout__column--2-3 display-block">
-        @sections.map { case Section(name, status, hasChanged, call) =>
-            @section(name, status, hasChanged, call)
+        @allSections.sortBy(section => Messages(s"progress.${section.name}.name")).map { s =>
+            @{ if(sections.contains(s)) section(s.name, s.status, s.hasChanged, s.call) else submissionSection(s.name, s.status, s.call)}
         }
```
slavvco: this will for example put ASP on top, and will mix the order of mandatory and optional sections, is that what we want to achieve?

owen-opencast: I've removed the ordering of sections - it looks like the default order is the order they're added (hence mandatory sections appearing first). All sections are now displayed in the same list with default ordering, so new sections will be added at the bottom.

slavvco: looks ok 👍

test/controllers/RegistrationProgressControllerSpec.scala
```diff
@@ -483,8 +483,7 @@ class RegistrationProgressControllerSpec extends AmlsSpec
 
           val html = Jsoup.parse(contentAsString(result))
 
-          html.select("#new-sections-list").text() must include(Messages("progress.hvd.name"))
```
matthewharris1: Is there a reason this was removed?

matthewharris1: Disregard this

mostafasina:
QA passed

PR #998, opened 2019-04-03
Amls 4122 - Renewal WhichCurrencies Page Split

matthewharris1:
This is the Page split for the currency exchange renewal flow. Youll find model, views, controllers and spec changes in both msb areas and renewal areas. The job was to split the WhichCurrencies page into 3 separate pages to fit standards and slot those pages in the right place in the renewal flow.

## How Has This Been Tested?
Manual Testing
Unit Testing
(Acceptance Testing but there are a few tests that need URL changes and button ID changes)

^ The UsesForeignCurrenciesYes / No buttons are now -true -false compared to the previous -yes -no label

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/models/renewal/WhichCurrencies.scala
```diff
@@ -16,231 +16,139 @@
...
+
+  def moneySources(p: MoneySources): WhichCurrencies =
+    this.copy(moneySources = Some(p))
+}
 
-  type MoneySourceValidation = (Option[BankMoneySource], Option[WholesalerMoneySource], Option[Boolean])
-  type WhichCurrenciesValidation = (Option[Boolean], Option[BankMoneySource], Option[WholesalerMoneySource], Option[Boolean])
+object WhichCurrencies {
 
-  private val emptyToNone: String => Option[String] = { x =>
```
samzaroughmrc: was there a reason to make the function object public, when it was private?

matthewharris1: Not really but can you explain how making it private would change anything please?

samzaroughmrc: we can have a session outside this PR about the impact of private/public on programs design.

app/controllers/renewal/UsesForeignCurrenciesController.scala
```diff
@@ -0,0 +1,102 @@
...
+        oldRenewal.whichCurrencies(newWc)
+      }
+      case _ => oldRenewal
+    }
+  }
+
+  def routing(msbServices: Set[BusinessMatchingMsbService],
+              edit: Boolean,
+              data: UsesForeignCurrencies)(implicit hc: HeaderCarrier, auth: AuthContext) = {
+      if (data == UsesForeignCurrenciesYes) {
```
ajchilts: Formatting?

app/models/renewal/WhichCurrencies.scala
ajchilts: Private means it's only accessible within this class. Taking away private implies it can be called from outside of the class by entities that instantiate this class. Its commonly used incorrectly. I would suggest if the method was private before and it does not need to be made public then keep it private. What is the justification for removing the private keyword? Does something fail if this val is private?

matthewharris1: Nothing fails I know the difference between public and private that wasn't my question. I was simply asking how it would affect this specific function. 

The method can be called outside of this class but it isn't, does not having the private keyword affect compilation times or something? I feel as though this keyword would only matter if someone tried to call this method outside this class which is most likely not going to happen as it's simply pulling a value from an option. 

I am happy to change this I am simply curious as to why its needed.

app/controllers/renewal/UsesForeignCurrenciesController.scala
matthewharris1: Done

slavvco:
No need for any changes or re-review, run of AC with encryption on/off is passing, just merged with master to have most updated version.

PR #1003, opened 2019-04-09
Amls 4880 - fix to supervision/msb sections after page split

slavvco:
In supervision and msb sections (page splits) when encryption is on the saving of data takes longer. Therefore value returned from save method on dataCacheConnector may not be available when doing redirect. In this solution is implemented separate cache read before redirect which solves the issue.
It's not possible to replicate this issue with encryption off.

## Related / Dependant PRs?

- none

## How Has This Been Tested?

- unit tests
- manual tests
- regression run

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/controllers/msb/MoneySourcesController.scala
```diff
@@ -64,17 +64,19 @@ class MoneySourcesController @Inject()(val authConnector: AuthConnector,
           case ValidForm(_, data: MoneySources) =>
             for {
               msb <- dataCacheConnector.fetch[MoneyServiceBusiness](MoneyServiceBusiness.key)
-              cache <- dataCacheConnector.save[MoneyServiceBusiness](MoneyServiceBusiness.key,
-                updateMoneySources(msb, data))
+              _ <- dataCacheConnector.save[MoneyServiceBusiness](MoneyServiceBusiness.key, updateMoneySources(msb, data))
+              maybeCache <- dataCacheConnector.fetchAll
+              cache <- Future.successful(maybeCache)
```
ajchilts: Should we map over here instead of calling future.successful? There are examples in the code where we map over a cache fetch.

app/controllers/supervision/AnotherBodyController.scala
```diff
@@ -29,59 +29,66 @@ import views.html.supervision.another_body
...
+      implicit request =>
+        Form2[AnotherBody](request.body) match {
+          case f: InvalidForm =>
+            Future.successful(BadRequest(another_body(f, edit)))
+          case ValidForm(_, data: AnotherBody) =>
+            for {
+              supervision <- dataCacheConnector.fetch[Supervision](Supervision.key)
+              _ <- dataCacheConnector.save[Supervision](Supervision.key, updateData(supervision, data))
+              maybeCache <- dataCacheConnector.fetchAll
+              cache <- Future.successful(maybeCache)
```
ajchilts: As above. Should we map over here instead of calling future.successful? There are examples in the code where we map over a cache fetch.

app/controllers/supervision/AnotherBodyController.scala
```diff
@@ -29,59 +29,66 @@ import views.html.supervision.another_body
...
     supervision.anotherBody(updatedAnotherBody).copy(hasAccepted = true)
   }
 
-  private def redirectTo(edit: Boolean, cache: CacheMap)(implicit authContext: AuthContext, headerCarrier: HeaderCarrier) = {
+  private def redirectTo(edit: Boolean, maybeCache: Option[CacheMap])(implicit authContext: AuthContext, headerCarrier: HeaderCarrier) = {
 
     import utils.ControllerHelper.{anotherBodyComplete, isAnotherBodyYes, supervisionComplete}
 
-    val anotherBody = anotherBodyComplete(cache)
+    maybeCache match {
```
ajchilts: If we implement this in the controllers as mentioned above and pass the cache in do we need to do this here?

app/controllers/supervision/SupervisionEndReasonsController.scala
```diff
@@ -62,8 +62,10 @@ class SupervisionEndReasonsController @Inject()(val dataCacheConnector: DataCach
           case ValidForm(_, data) =>
             for {
               supervision: Option[Supervision] <- dataCacheConnector.fetch[Supervision](Supervision.key)
-              cache <- dataCacheConnector.save[Supervision](Supervision.key,
+              _ <- dataCacheConnector.save[Supervision](Supervision.key,
                 updateData(supervision, data))
+              maybeCache <- dataCacheConnector.fetchAll
+              cache <- Future.successful(maybeCache)
```
ajchilts: As above.

slavvco:
sorry for that, but I spotted three commented lines in the specs, so I removed them, that's why there's my commit after review, no other changes

PR #1015, opened 2019-04-15
Amls 4717

samzaroughmrc:
update content, URLs and remove negative contractions within the renewal flow

## Checklist

- [x] Build passing.
- [x] Unit test approach and implementation has been reviewed with QA (testers).
- [x] RELEASE NOTES ADDED.


matthewharris1:
First thing. Could you add a description and fill in the correct boxes as discussed a fortnite ish ago.

conf/renewal.routes
```diff
@@ -3,44 +3,44 @@ POST        /progress                                                         @c
 
 GET         /what-you-need                                                    @controllers.renewal.WhatYouNeedController.get()
 
-GET         /confirm-activities                                               @controllers.renewal.InvolvedInOtherController.get(edit: Boolean ?= false)
-POST        /confirm-activities                                               @controllers.renewal.InvolvedInOtherController.post(edit: Boolean ?= false)
+GET         /other-activities                                                 @controllers.renewal.InvolvedInOtherController.get(edit: Boolean ?= false)
```
matthewharris1: Whitespace after activities?

app/controllers/renewal/InvolvedInOtherController.scala
```diff
@@ -54,25 +54,19 @@ class InvolvedInOtherController @Inject()(
...
+
+    typesString.map {
+      case t =>
+        t.map(item => {
+        if (item.toLowerCase.startsWith("a") || item.toLowerCase.startsWith("e")) {
+          s"an ${item.toLowerCase}"
+        } else {
+          s"a ${item.toLowerCase}"
+        }
+        })
```
matthewharris1: In my personal opinion I think this method is a lot harder to read/understand than the previous method. We'd also have to come back to change this solution if we ever had to add a new section that began with another vowel other than E and A. It boils down to personal preference I guess but I personally prefer how it's written on the left. All that would need to be changed is the order of case statements which is what we have done in previous content tickets.

drewmiley: I would keep createBusinessTypeOrderedList within this file.
Does it need to be a `def`, or can we create it as a `val`? I would much prefer it to be a `val` if possible.

app/controllers/renewal/InvolvedInOtherController.scala
```diff
@@ -54,25 +54,19 @@ class InvolvedInOtherController @Inject()(
...
-        case MoneyServiceBusiness => Messages("businessmatching.registerservices.servicename.lbl.05")
-        case TrustAndCompanyServices => Messages("businessmatching.registerservices.servicename.lbl.06")
-        case TelephonePaymentService => Messages("businessmatching.registerservices.servicename.lbl.07")
+  private def businessTypes(activities: BusinessMatching): Option[List[String]] = {
+    val typesString: Option[List[String]] = activities.createBusinessTypeOrderedList
+
+    typesString.map {
+      case t =>
+        t.map(item => {
+        if (item.toLowerCase.startsWith("a") || item.toLowerCase.startsWith("e")) {
```
drewmiley: Extract as vowel inc. i, o, and u. Might need to be put in a shared util.

app/controllers/renewal/InvolvedInOtherController.scala
```diff
@@ -54,25 +54,19 @@ class InvolvedInOtherController @Inject()(
...
-        case EstateAgentBusinessService => Messages("businessmatching.registerservices.servicename.lbl.03")
-        case HighValueDealing => Messages("businessmatching.registerservices.servicename.lbl.04")
-        case MoneyServiceBusiness => Messages("businessmatching.registerservices.servicename.lbl.05")
-        case TrustAndCompanyServices => Messages("businessmatching.registerservices.servicename.lbl.06")
-        case TelephonePaymentService => Messages("businessmatching.registerservices.servicename.lbl.07")
+  private def businessTypes(activities: BusinessMatching): Option[List[String]] = {
+    val typesString: Option[List[String]] = activities.createBusinessTypeOrderedList
+
+    typesString.map {
+      case t =>
```
drewmiley: I think this means we can use flatMap?

app/controllers/renewal/InvolvedInOtherController.scala
drewmiley: Single point of return, Extract out the an/a logic and return `s"${ prefix } ${ item.toLowerCase }"`

drewmiley: Indent please

app/controllers/renewal/InvolvedInOtherController.scala
drewmiley: After discussion, I agree that createBusinessTypeOrderedList is in the right place.
I would rename it to something more readable and evocative of what it is achieving.

drewmiley:
Can we refactor https://github.com/hmrc/amls-frontend/blob/master/app/views/businessactivities/involved_in_other_name.scala.html
to use `alphabeticalBusinessTypes` possibly?

app/controllers/renewal/InvolvedInOtherController.scala
samzaroughmrc: @matthyatt1 @drewmiley refactoring is done

app/controllers/renewal/InvolvedInOtherController.scala
samzaroughmrc: @matthyatt1 @drewmiley refactoring is done

samzaroughmrc: @matthyatt1 @drewmiley refactoring is done

samzaroughmrc: @matthyatt1 @drewmiley refactoring is done

conf/messages
```diff
@@ -1130,13 +1130,13 @@ businessmatching.registerservices.other.title = What other service do you need t
...
-businessmatching.registerservices.servicename.lbl.03 = Estate agency services
-businessmatching.registerservices.servicename.lbl.04 = High value dealing
-businessmatching.registerservices.servicename.lbl.05 = Money service business activities
-businessmatching.registerservices.servicename.lbl.06 = Trust or company services
-businessmatching.registerservices.servicename.lbl.07 = Telephone, digital and IT payment services
+businessmatching.registerservices.servicename.lbl.01 = Accountancy service provider
+businessmatching.registerservices.servicename.lbl.02 = Bill payment service provider
+businessmatching.registerservices.servicename.lbl.03 = Estate agency business
+businessmatching.registerservices.servicename.lbl.04 = High value dealer
+businessmatching.registerservices.servicename.lbl.05 = Money service business
```
matthewharris1: So these labels have been changed within messages.

The change was specifically for the page "How much total net profit did your business make in the last 12 months, from the following activities?"

These labels are used a number of times throughout the service for instance
 AMLSTurnoverController
 businessMatching models etc.

Does the removal of "activities" apply to all of these places. Because as I see it the ticket was to change this in one place.

app/controllers/renewal/InvolvedInOtherController.scala
```diff
@@ -54,25 +54,19 @@ class InvolvedInOtherController @Inject()(
...
-      a.businessActivities.map {
-        case AccountancyServices => Messages("businessmatching.registerservices.servicename.lbl.01")
-        case BillPaymentServices => Messages("businessmatching.registerservices.servicename.lbl.02")
-        case EstateAgentBusinessService => Messages("businessmatching.registerservices.servicename.lbl.03")
-        case HighValueDealing => Messages("businessmatching.registerservices.servicename.lbl.04")
-        case MoneyServiceBusiness => Messages("businessmatching.registerservices.servicename.lbl.05")
-        case TrustAndCompanyServices => Messages("businessmatching.registerservices.servicename.lbl.06")
-        case TelephonePaymentService => Messages("businessmatching.registerservices.servicename.lbl.07")
+  private def businessTypes(activities: BusinessMatching): Option[List[String]] = {
+    val typesString: Option[List[String]] = activities.alphabeticalBusinessTypes
```
drewmiley: Not sure we need to define this as a val? We only use it once, and it might be clearer to leave it as is.

app/models/businessmatching/BusinessMatching.scala
```diff
@@ -107,6 +108,21 @@ case class BusinessMatching(
     case _ => false
   }
 
+  def alphabeticalBusinessTypes()(implicit message: Messages): Option[List[String]] = {
```
drewmiley: Can this be a `val`?

app/controllers/renewal/InvolvedInOtherController.scala
drewmiley: `activities.alphabeticalBusinessTypes.map...`

conf/messages
samzaroughmrc: its been discussed and agreed with Lisa and Jeoff that the above are the set that we will use through out the service.
if the impact is huge on the acceptance test, lets re-assess the impact

matthewharris1: There are 10 acceptance test failures due to this so ill get to work on them now.

app/models/businessmatching/BusinessMatching.scala
drewmiley: We had a look; does not seem to be an obvious way of doing this.

app/controllers/renewal/InvolvedInOtherController.scala
drewmiley: We lose some of the flexibility of having the option in this case. We have renamed the case variable however.

PR #1024, opened 2019-04-29
AMLS-4199 | RP Role in Business Page Split

owen-opencast:
Split the start date question from 'Roles in Business' page onto its own page.
Updated the content on the two pages, and on the 'detailed answers' page.
Added a bulleted list of roles on the 'start date' page when more than one role is selected.
The role is displayed inline when there is one role selected.

[See ticket here](https://jira.tools.tax.service.gov.uk/browse/AMLS-4199).

## How Has This Been Tested?

Manual testing and Unit tests.

Unit tests have been updated for each of the two pages to cover the controllers, views, models, and the logic for updating Positions (the composite model for the two pages). 

Also added testing for the conditional bullet list - positions are bulleted on the 'start date' page if there are more than one.

Acceptance tests will need updated - @Bongowhack was updating these.

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/controllers/responsiblepeople/PositionWithinBusinessStartDateController.scala
```diff
@@ -0,0 +1,124 @@
...
+                  case _ => Redirect(routes.SoleProprietorOfAnotherBusinessController.get(index, edit, flow))
+                }
+            }
+          } recoverWith {
+            case _: IndexOutOfBoundsException => Future.successful(NotFound(notFoundView))
+          }
+        }
+  }
+
+  private[controllers] def hasNominatedOfficer(rpSeqOption: Option[Seq[ResponsiblePerson]]): Boolean = {
```
samzaroughmrc: this might not need to change here and now. but IMHO, this is feature envy and could/should be part of the ResponsiblePeople object.

app/controllers/responsiblepeople/PositionWithinBusinessStartDateController.scala
```diff
@@ -0,0 +1,124 @@
...
+                  case _
+                    => NotFound(notFoundView)
+                }
+              }).getOrElse(NotFound(notFoundView))
+            }
+          case ValidForm(_, data) => {
+            for {
+              _ <- updateDataStrict[ResponsiblePerson](index) { rp => rp.positions match {
+                  case Some(x) => rp.positions(Positions.update(x, data))
+                  case None => throw new IllegalStateException("Positions does not exist, cannot update start date")
```
samzaroughmrc: was wondering, whats going to happen to the app after the exception is thrown

app/controllers/responsiblepeople/PositionWithinBusinessStartDateController.scala
owen-opencast: I've removed these methods from both Controllers and added to the ResponsiblePerson object.

app/controllers/responsiblepeople/PositionWithinBusinessStartDateController.scala
owen-opencast: I've added a logging statement as discussed. The exception is for completeness - the thinking behind it is that it's a state that shouldn't ever occur. To avoid calling get() on an Option, or routing to "Page not found" (which is done frequently elsewhere and should be removed, but that's another issue....), I thought it would be better to throw a descriptive exception and allow Play's error handling to route the user appropriately.

PR #1038, opened 2019-05-09
Amls 4838 - Bug fix: Validation error messages not linked to group on MSB 'Where do you send the largest amounts of money' page

slavvco:
In this pull request there are implemented changes to SendTheLargestAmountsOfMoney model in order to display validation messages in similar way like it was done for MostTransactions. Updates to model, view, controller, specs and conversion methods. Loaded applications saved before changes and submitted them without problem, looks like there are no issues with compatibility.

## Related / Dependant PRs?

- PR in acceptance tests for that

## How Has This Been Tested?

- unit tests
- manual checks
- regression

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.

### DIFF
--- a/app/models/moneyservicebusiness/SendTheLargestAmountsOfMoney.scala
+++ b/app/models/moneyservicebusiness/SendTheLargestAmountsOfMoney.scala
@@ -73,6 +73,7 @@ private sealed trait SendTheLargestAmountsOfMoney0 {
       case (a, Some(b), Some(c)) => SendTheLargestAmountsOfMoney(Seq(a, b, c))
       case (a, Some(b), None) => SendTheLargestAmountsOfMoney(Seq(a, b))
       case (a, None, None) => SendTheLargestAmountsOfMoney(Seq(a))
+      case (_, _, _) => SendTheLargestAmountsOfMoney(Seq())
       }
     }
 
@@ -87,6 +88,7 @@ private sealed trait SendTheLargestAmountsOfMoney0 {
       case Seq(a, b, c) => Json.obj("country_1" -> a, "country_2" -> b, "country_3" -> c)
       case Seq(a, b) => Json.obj("country_1" -> a, "country_2" -> b)
       case Seq(a) => Json.obj("country_1" -> a)
+      case _ => Json.obj()
     }
   }
 }
--- a/app/models/moneyservicebusiness/SendTheLargestAmountsOfMoney.scala
+++ b/app/models/moneyservicebusiness/SendTheLargestAmountsOfMoney.scala
@@ -16,46 +16,87 @@
 
 package models.moneyservicebusiness
 
-import models.Country
-import models.FormTypes._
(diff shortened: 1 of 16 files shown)


app/models/moneyservicebusiness/SendTheLargestAmountsOfMoney.scala
```diff
@@ -16,46 +16,89 @@
...
-      )
+  val formW: Write[SendTheLargestAmountsOfMoney, UrlFormEncoded] = {
+    import jto.validation.forms.Writes._
+    import utils.MappingUtils.spm
+    implicitly
+  }
+
+  val jsonW = Writes[SendTheLargestAmountsOfMoney] { lom =>
+    lom.countries match {
+      case Seq(a, b, c) => Json.obj("country_1" -> a, "country_2" -> b, "country_3" -> c)
```
drewmiley: I think this is the best way as it is clearer about its intention

app/services/LandingService.scala
```diff
@@ -283,7 +283,11 @@ class LandingService @Inject() (
             (_.transactionsInNext12Months.map(t => TransactionsInLast12Months(t.txnAmount))),
 
           sendTheLargestAmountsOfMoney = viewResponse.msbSection.fold[Option[SendTheLargestAmountsOfMoney]](None)
-            (_.sendTheLargestAmountsOfMoney.map(s => SendTheLargestAmountsOfMoney(s.country_1, s.country_2, s.country_3))),
+            (_.sendTheLargestAmountsOfMoney.map (s => s.countries match {
+              case Seq(a, b, c) => SendTheLargestAmountsOfMoney(a, Some(b), Some(c))
```
drewmiley: Could maybe do something with .head and .tail

drewmiley: But i'm not massively keen

app/views/msb/send_largest_amounts_of_money.scala.html
```diff
@@ -38,18 +38,24 @@
 
         @fieldset(
             legend = "msb.send.the.largest.amounts.of.money.title",
-            f = f("countries"),
+            f = f("largestAmountsOfMoney-fieldset"),
             legendHidden = true,
-            panel = false
+            panel = false,
+            classes = Seq("inline")
```
drewmiley: Do we need this?

app/services/LandingService.scala
slavvco: if you don't mind I'll leave this as there are only max three elements in seq

app/views/msb/send_largest_amounts_of_money.scala.html
slavvco: nope we don't :) I'll remove

app/services/LandingService.scala
drewmiley: Agreed

PR #1126, opened 2019-07-19
Amls 4998

ajchilts:
Initial Preparation - Moving to the new Auth Client FE including ew AuthAction with required retrievals. 

## Related / Dependant PRs?

N/A

## How Has This Been Tested?

Unit and Local with one sample controller migration (backed out as to be done in separate ticket) + regression pack. 

In order to achieve this I need to also do some work to make the OID in the cache connector optional such that we can proceed once it is deprecated and no longer exists. 

New AuthAction added and ready for use. Can be extended if more retrievals needed. 

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/connectors/cache/MongoCacheConnector.scala
```diff
@@ -56,7 +58,14 @@ class MongoCacheConnector @Inject()(cacheClientFactory: MongoCacheClientFactory,
...
       credId => {
-        mongoCache.createOrUpdate(credId, authContext.user.oid, data, key).map(toCacheMap)
+        mongoCache.createOrUpdate(credId, Some(authContext.user.oid), data, key).map(toCacheMap)
+      }
+    }
+  }
+  def save[T](id: String, key: String, data: T)(implicit hc: HeaderCarrier, format: Format[T]): Future[CacheMap] = {
+    authConnector.getCredId flatMap {
+      credId => {
+        mongoCache.createOrUpdate(credId, None, data, key).map(toCacheMap)
```
slavvco: here we are asking authConnector for credId, but id is credId?

ajchilts: Will take a 👀

PR #1161, opened 2019-08-09
Remove question is your company registered for corporatrion tax from …

ajchilts:
Remove question is your company registered for corporatrion tax from flow as we should not give user oportunity to override this from business matching.

Fix was not as simple as removing the question, page and models etc. 

User enters UTR in BM
 - Its then placed in the BM section in the mongo cache
 - When the user fills out the Business Details section we copy the UTR over from BM

 - The UTR from Business Details is used to populate the UTR in DES schema NOT the UTR from BM!

 - API5 responds with the UTR in the DES schema and we translate into Business Details (NOT BM!)

 - The value is then synced with ETMP via DES using the Business Details section. 

The above means if we wanted to remove the UTR from business details we would have to rewrite all of the parsing and translation logic between AMLS and DES to use BM for UTR and not Business Details. This is a large bit of work that would require changes both in AMLS FE and BE and would required a rewrite if the JSON reads and writes for multiple objects. 

To get around this I have deleted the page only; leaving the corporation tax controller in place. We no longer show the page. If the UTR exists in BM section of the cache it means we are pre app and we should write it to Business Details and skip the question. If we do not have the UTR in BM then we are not pre app and we need do nothing other than redirect as API5 will keep this field up to date for us. 

## Related / Dependant PRs?

N/A

## How Has This Been Tested?

Unit + Local + Regression Pack

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/controllers/businessdetails/CorporationTaxRegisteredController.scala
```diff
@@ -47,33 +48,25 @@ class CorporationTaxRegisteredController @Inject () (
...
+                details <- OptionT.fromOption[Future](bm.reviewDetails)
+                // Only update Business Details from Business Matching where it exists; after that its maintained by API5
+                _ <- if (details.utr.isDefined) {
+                  updateCache(cache, CorporationTaxRegisteredYes(details.utr.getOrElse(
+                    throw new Exception("[CorporationTaxRegisteredController][get]: Could not retrieve UTR from Business Matching")
+                  )))
+                } else {
+                  OptionT.fromOption[Future](Some(cache))
+                }
+              } yield getRedirectLocation(edit)) getOrElse InternalServerError("Could not route from CorporationTaxRegisteredController")
```
owen-opencast: User should never be able to edit this page. Could use Redirect(routes.ConfirmRegisteredOfficeController.get()) directly here and remove the getOrElse?

ajchilts: Totally take your point on the edit however; the get or else is an action based on cache retrieval and redirect and not just redirect so I think it will need to stay as belt and braces. It handles the eventuality that something might not complete successfully in the cache related futures. Should never be hit but worth logging if it is.

PR #1176, opened 2019-08-21
AMLS-5246 | Fixed misplaced errors in "Which currencies..."

owen-opencast:
Fixed misplaced errors when an empty field is submitted preceding fields with data. The form validation strips out empty fields, as users can enter fewer than 3 currencies. As the empty fields are kept in the form data used to repopulate the page, the errors misalign. The solution strips out empty fields from the data used to repopulate the page so the data and errors align correctly.

NB: This fix will be required if the same issue occurs on other currency pages, as it is applied on a per controller basis.

## Related / Dependant PRs?

This is the same solution as for AMLS-5116 | BranchesOrAgentsWhichCountriesController, but with a small tweak to the logic due to differences in the form data.

## How Has This Been Tested?

Manual testing, spec tests, acceptance tests.

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/controllers/msb/WhichCurrenciesController.scala
```diff
@@ -50,21 +50,31 @@ class WhichCurrenciesController @Inject() (val authConnector: AuthConnector,
...
-                  }
+              } yield edit match {
+                case true => Redirect(routes.SummaryController.get())
+                case _ => Redirect(routes.UsesForeignCurrenciesController.get())
+              }
       }
     }
   }
 
+  // Validation removes empty fields. This removes them from the form data, so the errors align correctly with the fields.
```
matthewharris1: Not sure we need the comment here. Goes against centre guidelines, Code should be explanatory when read which it is. other than that sound job :+1:

app/controllers/msb/WhichCurrenciesController.scala
```diff
@@ -50,21 +50,31 @@ class WhichCurrenciesController @Inject() (val authConnector: AuthConnector,
...
+  def removeEmptyFields(f: InvalidForm): InvalidForm = {
+    val csrfToken = f.data.head
+    val fieldsWithData:Map[String, Seq[String]] = f.data
+      .filter(field => field._1.contains("currencies"))
+      .filter(field => field._2.exists(s => s.nonEmpty))
+      .zipWithIndex
+      .map((tuple: ((String, Seq[String]), Int)) => (tuple._1._1.replaceFirst("[\\d]", s"${tuple._2}"), tuple._1._2))
+    f.copy(data = fieldsWithData + csrfToken)
+  }
+
```
slavvco: Similar function exists in other controller, is there a way to generalise it and reuse? Can we have better naming here and split it if possible as it's hard to read?

app/controllers/msb/WhichCurrenciesController.scala
owen-opencast: I disagree about comments, I think they can be useful for many reasons - for example, providing context. Reading code provides what and how, but it might not be self-evident why. 

If a comment saves someone valuable time wondering why a piece of code is there, removing it, breaking things, having to re-add it (potentially in a few years, after every current dev has left), then I think they're valid.

That being said, if there are centre guidelines you can point me to that say "DO NOT USE COMMENTS", then I can remove it. But I still disagree :)

app/controllers/msb/WhichCurrenciesController.scala
owen-opencast: I'll look into this - they differ in the .map() as the forms are slightly different in structure but it'll be do-able.

app/controllers/msb/WhichCurrenciesController.scala
matthewharris1: We’ve discussed the use of comments multiple times in the past and came to the agreement not to use them. It is centre guidelines I’m not sure where it’s written but I’ve been told by nick wolfie etc not to do it. If you have a big issue removing it maybe ask The rest of the team about this

owen-opencast: I don't have a big issue over this particular comment. We've discussed comments before, but I don't think the issue has ever been definitely settled, as I said, there are valid reasons for using them. If it is a centre guideline, I'd like to see the guidelines, if only to know how else I might inadvertently be contravening the centre's house-style.

owen-opencast:
I've made the requested changes. I've extracted a method that both fixed pages now use - the same method can now be used to fix any other affected pages.

owen-opencast:
Oh, and added some spec tests for the extracted method.

app/controllers/msb/WhichCurrenciesController.scala
matthewharris1: Thanks for removing :+1:

PR #1193, opened 2019-09-19
Amls 5299

ajchilts:
AMP FE cache repository switch out to read write to AMLS cache

## Related / Dependant PRs?

https://github.com/hmrc/amls-art-market-participant-frontend/pull/14

## How Has This Been Tested?

Unit + Local

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/models/amp/Amp.scala
```diff
@@ -26,7 +26,7 @@ import typeclasses.MongoKey
 import uk.gov.hmrc.http.cache.client.CacheMap
 import uk.gov.hmrc.play.config.ServicesConfig
 
-final case class Amp(id: String,
+final case class Amp(_id: String,
```
owen-opencast: Might be worth naming this credId for clarity?

conf/messages
```diff
@@ -728,6 +728,9 @@ progress.tradingpremises.time = 3 and 8
 progress.asp.name = accountancy service provider
 progress.asp.description = Registration information specifically for Accountancy Service Providers.
 progress.asp.time = 2 and 4
+progress.amp.name = art market participant
+progress.amp.description = Registration information specifically for Art Market Participants.
+progress.amp.time = ? and ?
```
owen-opencast: I don't think description or time are used anywhere, can be removed

test/controllers/amp/AmpControllerSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+  val completeData = Json.obj(
+    "typeOfParticipant"             -> Seq("artGalleryOwner"),
+    "boughtOrSoldOverThreshold"     -> true,
+    "dateTransactionOverThreshold"  -> LocalDate.now,
+    "identifyLinkedTransactions"    -> true,
+    "percentageExpectedTurnover"    -> "fortyOneToSixty"
+  )
+
+  val completeJson = Json.obj(
+    "_id"            -> "someid",
```
owen-opencast: _credId?

app/models/amp/Amp.scala
ajchilts: I wanted to keep this exactly as it is stored in AMP mongo for the time being in case we wanted to switch back. 

We may have a discussion and do a refactoring exercise to remove these additional fields however; that could be its own ticket if we decide we want to go that way.

test/controllers/amp/AmpControllerSpec.scala
ajchilts: As above...

app/services/amp/AmpService.scala
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  def set(credId: String, body: JsValue)(implicit hc: HeaderCarrier) = {
+    val jsonObject: JsObject = body.as[JsObject]
+    val ampData = jsonObject.value("data").as[JsObject]
+
+    for {
+      existing <- cacheConnector.fetch[Amp](credId, Amp.key)
+      result   <- cacheConnector.save[Amp](credId, Amp.key, existing.getOrElse(Amp(credId)).data(ampData))
+    } yield result
```
slavvco: Let's say someone has Amp data already stored... and is updating something. Will this set method update the data? It looks like it's saving the data only if there is no existing data?

app/services/amp/AmpService.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class AmpService @Inject()(cacheConnector: DataCacheConnector)
+                          (implicit ec: ExecutionContext){
+
+  def get(credId: String)(implicit hc: HeaderCarrier): Future[Option[JsValue]] = {
+    for {
+      r <- cacheConnector.fetch[Amp](credId, Amp.key)
+    } yield r match {
```
slavvco: what's "r"? more meaningful name?

app/services/amp/AmpService.scala
ajchilts: The data method is called after the 'get or else' so it's always called either way.

slavvco: Ok. It wasn't clear for me what data method is doing.

PR #1203, opened 2019-10-04
AMLS-5333 | Update AMP models

owen-opencast:
Updated AMP models to connect up to back-end

## How Has This Been Tested?

Manual testing and running spec tests

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/services/amp/AmpCacheService.scala
```diff
@@ -42,7 +42,7 @@ class AmpCacheService @Inject()(cacheConnector: DataCacheConnector)
 
     for {
       existing <- cacheConnector.fetch[Amp](credId, Amp.key)
-      result   <- cacheConnector.save[Amp](credId, Amp.key, existing.getOrElse(Amp(credId)).data(ampData))
+      result   <- cacheConnector.save[Amp](credId, Amp.key, existing.map(amp=> amp.data(ampData)).getOrElse(Amp().data(ampData)))
```
ajchilts: Does this hold if existing is null? 

Used to be `existing.getOrElse(Amp(credId)).data(ampData))` to cater for existing being null. What happens if we map over null and try and set data? Also the removed line is less code? If there is a valid reason for changing this then fair enough.

conf/messages
```diff
@@ -610,7 +610,7 @@ status.submissionreadyforreview.description.5 = Contact HMRC if you have a quest
 status.submissionreadyforreview.description.5.link = the Money Laundering Regulations
 status.submissionreadyforreview.description.link = pay your fees now
 status.submissionreadyforreview.nofee.description = You have submitted your application. You have 28 days to pay the fees from the date you submitted.
-status.submissionreadyforreview.nofee.description.link = If you have not already paid
+status.subtumissionreadyforreview.nofee.description.link = If you have not already paid
```
ajchilts: Is this an intentional change?

owen-opencast: No. No it was not...

PR #1264, opened 2019-11-26
Amls 4881 - New notifications layout

slavvco:
As above. More in the ticket.

## Related / Dependant PRs?

- none

## How Has This Been Tested?

- manually

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.

### DIFF
--- a/public/amls.css
+++ b/public/amls.css
@@ -28,6 +28,10 @@ a.block-link {
   margin: 0;
 }
 
+.heading-medium.no-bottom-margin {
+  margin-bottom: 0;
+}
+
 /* new-style page headers */
 
 .page-header {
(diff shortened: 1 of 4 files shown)


public/amls.css
```diff
@@ -1353,6 +1353,11 @@ table.notifications thead td {
   text-decoration: underline;
 }
 
+.message-inbox__list
```
tempertemper: Is this line needed? It should have an underline until it has focus (I'm flying blind here! Might be worth my looking over your shoulder tomorrow morning)

slavvco: it works as you wanted I believe, but we can have a look together when you're in

PR #1299, opened 2020-01-10
Amls 5587 make pay api change revertible

ajchilts:
Amls 5587 make pay api change revertible by changing Payment description field to an option.

## Related / Dependant PRs?

https://github.com/hmrc/amls/pull/282

## How Has This Been Tested?

Local + Regression

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.

### DIFF
--- a/test/generators/PaymentGenerator.scala
+++ b/test/generators/PaymentGenerator.scala
@@ -54,7 +54,7 @@ trait PaymentGenerator extends BaseGenerator with AmlsReferenceNumberGenerator {
     amlsRefNo,
     safeId,
     ref,
-    desc,
+    Some(desc),
     amountInPence,
     paymentStatus,
     now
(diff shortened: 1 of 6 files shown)


ajchilts:
@slavvco @Bongowhack  - Lets have a discussion and see if we feel we should do this.

test/generators/PaymentGenerator.scala
```diff
@@ -54,7 +54,7 @@ trait PaymentGenerator extends BaseGenerator with AmlsReferenceNumberGenerator {
     amlsRefNo,
     safeId,
     ref,
-    desc,
+    Some(desc),
```
slavvco: so payment generator will always generate some description? what about no-description?

ajchilts: @slavvco - As this code is on the old API there will always be a description and this is just a test generator. The next set of PRs to cover the API endpoint change will add some tests that cover no description. Make sense?

PR #1320, opened 2020-01-31
Amls 5647 | New status page 'Application pending'

slavvco:
Implemented new design of Application pending page. Renamed some files and views. Added functionality to count and display the number of unread messages.

## Related / Dependant PRs?

- none

## How Has This Been Tested?

- manually
- unit tests

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


slavvco:
test this please

slavvco:
test this please

app/controllers/StatusController.scala
```diff
@@ -251,13 +256,24 @@ class StatusController @Inject()(val landingService: LandingService,
...
+      case (true, _, true) | (_, true, true) => trade_information()
       case (_, _, _) => throw new MatchError("Could not match activities against given options.")
     }
+
+  def countUnreadNotifications(amlsRefNo: Option[String], safeId: Option[String], accountTypeId: (String, String))(implicit headerCarrier: HeaderCarrier) = {
+    val notifications = (amlsRefNo, safeId) match {
+      case (Some(ref), _) => notificationConnector.fetchAllByAmlsRegNo(ref, accountTypeId)
+      case (None, Some(id)) => notificationConnector.fetchAllBySafeId(id, accountTypeId)
+      case (None, None) => Future.successful(Seq())
+      case (_, _) => throw new MatchError("Could not match amls ref number or safe id against given conditions.")
```
ajchilts: Might be less disruptive to return a zero here rather than raise an exception? An exception will cause a 'Please try again later'. If this is non recoverable it will block the user from accessing the service for an unread notification count. Maybe we could log a warning so we can trace in Kibana however; return a zero and allow them to carry on where we can't determine the count?

test/controllers/StatusControllerSpec.scala
```diff
@@ -805,14 +811,114 @@ class StatusControllerSpec extends AmlsSpec with PaymentGenerator with PrivateMe
...
+        result returns 0
+      }
+
+      "throw an exception if no match found" in new Fixture with FutureAssertions {
+
+        val result = intercept[MatchError] {
+          controller.countUnreadNotifications(null, null, ("", ""))
+        }
+
+        result.getMessage() must include("Could not match amls ref number or safe id against given conditions.")
```
ajchilts: I think it might be safer to log a warn and return zero. It would then be non blocking for the user for the sake of a message count.

app/controllers/StatusController.scala
slavvco: Ok, thanks for the review, I'll address above matter in the morning.

slavvco: Done, default case will be empty seq, so it will return 0. Removed the MatchError and corresponding test.

test/controllers/StatusControllerSpec.scala
slavvco: Done :)

PR #1343, opened 2020-02-13
Prevent save on position in business where other is selected and only…

ajchilts:
Prevent save on position in business where other is selected and only white space is provided

## Related / Dependant PRs?

N/A

## How Has This Been Tested?

Local + Unit

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/models/responsiblepeople/PositionWithinBusiness.scala
```diff
@@ -118,6 +118,8 @@ object PositionWithinBusiness
   private[responsiblepeople] val fullySpecifiedRule = Rule[(Set[String], Option[String]), Set[PositionWithinBusiness]]  {
     case (s, None) if s.contains("other") =>
```
matthewharris1: Can this None be changed to _? Then we don't need the second "duplicate" case

PR #1357, opened 2020-02-20
Amls 5445 - MSB error changes

matthewharris1:
MSB section error updates alongside 1 H1 change.
## How Has This Been Tested?
Manual
Unit

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/models/moneyservicebusiness/MoneySources.scala
```diff
@@ -37,11 +37,18 @@ object MoneySources {
   import jto.validation.forms.Rules._
   import utils.MappingUtils.Implicits._
 
-  private def nameType(fieldName: String) = {
+  private def bankNameType(fieldName: String) = {
```
slavvco: because this def is not being reused anymore you can get rid of the parameter and change it to val

app/models/moneyservicebusiness/MoneySources.scala
```diff
@@ -37,11 +37,18 @@ object MoneySources {
...
     notEmptyStrip andThen
       minLength(1).withMessage(s"error.invalid.msb.wc.$fieldName") andThen
-      maxLength(140).withMessage("error.invalid.maxlength.140") andThen
-      basicPunctuationPattern()
+      maxLength(140).withMessage("error.maxlength.msb.wc.bankNames") andThen
+      basicPunctuationPattern("error.format.msb.wc.banknames")
+  }
+
+  private def wholeSalerNameType(fieldName: String) = {
```
slavvco: same as above

app/models/moneyservicebusiness/MoneySources.scala
matthewharris1: Good spot changed :)

app/models/moneyservicebusiness/MoneySources.scala
matthewharris1: ^

PR #1373, opened 2020-03-06
Amls 5796 - Cache validation and refresh from API5 if no cache when hitting status controller

ajchilts:
We've identified what's causing a number of errors we didn't understand why were happening.

The errors;
```
Cannot get business matching or renewal date
Unable to redirect from Turnover page
Unable to retrieve SafeID from reviewDetails
Could not match activities against given options
```

I believe are being caused by the same problem.

When investigating in local dev I found that you could bypass the landing controller (bit that checks if the cache should be reloaded from API5) by bookmarking the application status page and going to that bookmark while already signed in to GG in an active session.

If I login in one window; then close the window and drop the cache (replicate the 30 day expiry); then hit my application status book mark the service does not reload the cache. The page renders from API9 as expected and prompts the user to start their renewal. The user starts the renewal by answering the first question. This then creates a cache for the user however; it has only the renewal section and no other sections. The system then throws the 'unable to redirect from turnover page' and the user drops back to the application status page. If the user attempts to go to the about your business page then they get the 'Unable to retrieve SafeID from reviewDetails' error and the 'Could not match activities against given options' error.

I discussed this scenario with Mike Pratt who then proved the theory in the QA environment. In addition to proving this in QA Mike Pratt was able to trace a few of the live user journeys through Splunk and was able to verify this scenario to be the case in production.

Fix: The fix is to modify the GET method in the application status page to first fetch the cache. If the cache is present then carry on as normal. If the cache is not present then reload from API5 before proceeding. This means the user should always have the data required to proceed.

## Related / Dependant PRs?

N/A

## How Has This Been Tested?

Unit + Local

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.

### DIFF
--- a/app/controllers/StatusController.scala
+++ b/app/controllers/StatusController.scala
@@ -55,9 +55,9 @@ class StatusController @Inject()(val landingService: LandingService,
 
   def get(fromDuplicateSubmission: Boolean = false) = authAction.async {
     implicit request =>
-
       request.amlsRefNumber match {
         case Some(amlsRefNo) =>
+          // MUST ensure we have a cache here! If the cache has expired reload from API5 as it must be populated to succeed.
           renewalService.isCachePresent(request.credId) flatMap {
             case true => getPage(
               request.amlsRefNumber,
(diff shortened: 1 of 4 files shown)


ajchilts:
An alternative fix might be simply to redirect to the landing controller if there is no cache rather than loading the cache from API5. I will explore this as its likely safer.

PR #1387, opened 2020-03-17
AMLS-5777 | New notification templates for secure comms

matthewharris1:
This branch needs an update in notification config to makes sure that the new version is being used (v4m0)
This also needs updates in app-config  libraries before deploying.

## How Has This Been Tested?
Manually

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


ajchilts:
@matthewharris1 - do we need to prep some config for this also? I.e. new version template config?

ajchilts:
> @matthewharris1 - do we need to prep some config for this also? I.e. new version template config?

Oh - just actually read the PR. :-) 

We should get that config ready to go.

matthewharris1:
@ajchilts Will get on to that now :)

PR #1391, opened 2020-03-19
Amls 5825 | Bug fix: duplicate bank accounts when navigating backward to memorable name page

slavvco:
How to reproduce the bug is described in the ticket, but basically when entering memorable name for bank account the system creates entry in the DB. When later navigating forward everything works fine, but when using backward navigation and going back to name page the data is not being played back in the form. In this case the user might enter again the same name. When posting this new entry is created again. Later when going through the all questions and answering them, accepting the CYA page on Your bank accounts we can see one complete account one incomplete with the same names. Basically it's not a big issue because user can remove the duplicate, but in the ideal world we would like to avoid unclear situations like this.

Core problem here is that when navigating backwards previous page has no information about the details of the next one - back buttons are using browser history, so going back we simply make call to the past and that past is not holding any information about the future - like in most cases in life :)

So again going back we simply reply the url that was stored in browser history, this url doesn't have the information about the database entry created previously and here we have a small issue.

The solution is to somehow pass the information about new entry in such a way that it can be read easily in the future and past, as well as deleted easily.

in this case, after considering different other solution, I've decided then to pass extra parameter with item index as session parameter, as that is quite straightforward solution. This parameter allows to load proper page with data/or no when navigating backwards. It's always cleared out (if exists) when going through WYN page or when accepting CYA page.

This solution is not perfect, but will minimise the possibility of creating duplicates. It needs testing in the environments.

## Related / Dependant PRs?

- none

## How Has This Been Tested?

- manually
- unit tests

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


slavvco:
@ajchilts - in the commit above I've added removing the itemIndex from session in one more controller. This will ensure that after completion of this section there will be no itemIndex parameter in  the session at all.

PR #1437, opened 2020-04-24
Amls 5894 | Fix for empty countries lists in renewal

slavvco:
This will fix the problem with empty lists being send while renewing the application.

Initial thoughts were that we are creating empty lists in frontend, but after small investigation I've found that we are missing one question when converting renewal data to subscription request. This was the problem. Because of that the AMLS while creating the request was using not updated data for this question and was creating empty lists in result.

## Related / Dependant PRs?

- none

## How Has This Been Tested?

- manually
- regression
- unit tests

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


slavvco:
@ajchilts I've checked this and the code how it is right now is not writing empty lists in Mongo :)

PR #1457, opened 2020-05-07
Amls 5819 | CYA changes in MSB

slavvco:
As above.

## Related / Dependant PRs?

- none

## How Has This Been Tested?

- manually
- unit tests

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.


app/models/moneyservicebusiness/MoneySources.scala
```diff
@@ -31,7 +31,10 @@ case object CustomerMoneySource
 
 case class MoneySources(bankMoneySource: Option[BankMoneySource] = None,
                          wholesalerMoneySource: Option[WholesalerMoneySource] = None,
-                         customerMoneySource: Option[Boolean] = None)
+                         customerMoneySource: Option[Boolean] = None) {
+
+  def size = List(this.bankMoneySource, this.wholesalerMoneySource, this.customerMoneySource).flatten.size
```
ajchilts, line 36: This is a much nicer way of doing this. :-) 👍

slavvco:
Hi @ajchilts, latest commit is a fix to a bug existing in autocomplete component (https://github.com/alphagov/accessible-autocomplete/issues/432) in currencies. This solution is nothing new, was already implemented for countries in the past but seems like it must be done for currencies too. The bug: once selected and saved currency user comes back to remove currency - it does not work, because autocomplete is still holding old value. This piece of code fixes this behaviour.

PR #1463, opened 2020-05-13
Amls 5759 | Update H1 on Confirm the postcode page in pre-application and improve routing

slavvco:
In this PR there are significant changes to the flow from LandingController depending on the postcode/country code. There are also other changes, so please review carefully :D I've left some comments in this PR in a few places.

Testing:
1. Clone business-customer-stub
2. In EtmpController manipulate address as you like:
    val address = Address("23 High Street", "Park View", Some("Gloucester"), Some("Gloucestershire"), Some("NE98 1ZZ"), "GB")
3. Stop this microservice in service manager
4. Run cloned service on 9926
5. Load empty application with IR-SA/UTR/1111111111 preset
6. You will land on Check this is the business you want to register and it should display adress modified above
7. Continue with testing

## Related / Dependant PRs?

- none

## How Has This Been Tested?

- manually (using business-customer-stub)
- unit tests

## Checklist

- [x] Breaking change (fix or feature that would cause existing functionality to change).
- [x] Build passing.
- [x] RELEASE NOTES ADDED.

### DIFF
--- a/app/controllers/businessdetails/ConfirmRegisteredOfficeController.scala
+++ b/app/controllers/businessdetails/ConfirmRegisteredOfficeController.scala
@@ -23,7 +23,7 @@ import connectors.DataCacheConnector
 import controllers.{AmlsBaseController, CommonPlayDependencies}
 import forms.{EmptyForm, Form2, InvalidForm, ValidForm}
 import models.businesscustomer.Address
-import models.businessdetails.{BusinessDetails, ConfirmRegisteredOffice, RegisteredOffice, RegisteredOfficeUK}
+import models.businessdetails._
 import models.businessmatching.BusinessMatching
 import play.api.mvc.MessagesControllerComponents
 import utils.AuthAction
@@ -37,15 +37,26 @@ class ConfirmRegisteredOfficeController @Inject () (val dataCache: DataCacheConn
                                                     val ds: CommonPlayDependencies,
                                                     val cc: MessagesControllerComponents) extends AmlsBaseController(ds, cc) {
 
-
   def updateBMAddress(bm: BusinessMatching): Option[RegisteredOffice] = {
-    bm.reviewDetails.fold[Option[RegisteredOffice]](None)(dtls => Some(RegisteredOfficeUK(
-      dtls.businessAddress.line_1,
-      dtls.businessAddress.line_2,
-      dtls.businessAddress.line_3,
-      dtls.businessAddress.line_4,
-      dtls.businessAddress.postcode.getOrElse("")
-    )))
+    bm.reviewDetails.fold[Option[RegisteredOffice]](None)(dtls =>
+      if(dtls.businessAddress.postcode.isDefined) {
+        Some(RegisteredOfficeUK(
+          dtls.businessAddress.line_1,
+          dtls.businessAddress.line_2,
+          dtls.businessAddress.line_3,
+          dtls.businessAddress.line_4,
+          dtls.businessAddress.postcode.getOrElse("")
+        ))
+      } else {
+        Some(RegisteredOfficeNonUK(
+          dtls.businessAddress.line_1,
+          dtls.businessAddress.line_2,
+          dtls.businessAddress.line_3,
+          dtls.businessAddress.line_4,
+          dtls.businessAddress.country
+        ))
+      }
+    )
   }
 
   def getAddress(businessMatching: Future[Option[BusinessMatching]]): Future[Option[Address]] = {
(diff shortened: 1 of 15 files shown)


app/controllers/businessdetails/ConfirmRegisteredOfficeController.scala
```diff
@@ -37,15 +37,26 @@ class ConfirmRegisteredOfficeController @Inject () (val dataCache: DataCacheConn
...
+      } else {
+        Some(RegisteredOfficeNonUK(
+          dtls.businessAddress.line_1,
+          dtls.businessAddress.line_2,
+          dtls.businessAddress.line_3,
+          dtls.businessAddress.line_4,
+          dtls.businessAddress.country
+        ))
+      }
+    )
```
slavvco, line 59: routing was extended to allow progress to the service for customers with non-UK country code, therefore this change to extend the registered office address to accept the non-UK one

app/controllers/businessmatching/ConfirmPostCodeController.scala
```diff
@@ -43,7 +45,7 @@ class ConfirmPostCodeController @Inject()(authAction: AuthAction,
 
   def updateReviewDetails(reviewDetails: Option[ReviewDetails], postCodeModel: ConfirmPostcode): Option[models.businesscustomer.ReviewDetails] = {
     reviewDetails.fold[Option[ReviewDetails]](None) { dtls =>
-      val updatedAddr = dtls.businessAddress.copy(postcode = Some(postCodeModel.postCode))
+      val updatedAddr = dtls.businessAddress.copy(postcode = Some(postCodeModel.postCode), country = Country("United Kingdom", "GB"))
```
slavvco, line 48: this was changed because previously you could have in business details address with empty country Country("","") and valid UK postcode. This will now update not only postcode but also will set the country to UK if postcode will pass the validation

app/models/businesscustomer/ReviewDetails.scala
```diff
@@ -61,8 +61,13 @@ object ReviewDetails {
...
 
-  implicit def convert(addr: BusinessMatchingAddress): Address =
-    Address(addr.line_1, addr.line_2, addr.line_3, addr.line_4, addr.postcode, Country("", addr.country))
+  implicit def convert(addr: BusinessMatchingAddress): Address = {
+      val country = models.countries collectFirst {
+        case country@Country(_, code) if code == addr.country => country
+      } getOrElse Country("", addr.country)
+
+      Address(addr.line_1, addr.line_2, addr.line_3, addr.line_4, addr.postcode, country)
+    }
```
slavvco, line 70: this change was necessary because I've noticed that previous implementation was creating empty country objects in some instances => so this conversion wasn't all the time right. For example in BusinessMatchingAddress there is country as String (place for country code) => so original conversion for United Kingdom was creating: Country("", "GB") but it should be: Country("United Kingdom", "GB").

app/controllers/businessdetails/ConfirmRegisteredOfficeController.scala
```diff
@@ -37,15 +37,26 @@ class ConfirmRegisteredOfficeController @Inject () (val dataCache: DataCacheConn
...
   def updateBMAddress(bm: BusinessMatching): Option[RegisteredOffice] = {
-    bm.reviewDetails.fold[Option[RegisteredOffice]](None)(dtls => Some(RegisteredOfficeUK(
-      dtls.businessAddress.line_1,
-      dtls.businessAddress.line_2,
-      dtls.businessAddress.line_3,
-      dtls.businessAddress.line_4,
-      dtls.businessAddress.postcode.getOrElse("")
-    )))
+    bm.reviewDetails.fold[Option[RegisteredOffice]](None)(dtls =>
+      if(dtls.businessAddress.postcode.isDefined) {
```
ajchilts, line 42: Do we have a test that covers dtls being `None`? Just to make sure this line can handle that?

app/models/businesscustomer/ReviewDetails.scala
ajchilts: This sort of thing could have caused down stream issues. Good spot. 👍

app/controllers/businessdetails/ConfirmRegisteredOfficeController.scala
slavvco: Thanks, this was a valid point. Theoretically this should never happen that there is no review details but just in case, I've tightened condition on redirection to next page. Now if the address is none user will be redirected to address questions. There is also test for that.

slavvco:
@ajchilts last commit is to solve problem with None case when updating address.

PR #1481, opened 2020-06-01
AMLS-5933 | TP postcode fix for enrolments issue

ajchilts:

## Checklist

- [x] Build passing.
- [x] RELEASE NOTES ADDED.

### DIFF
--- a/app/services/SubmissionService.scala
+++ b/app/services/SubmissionService.scala
@@ -40,6 +40,7 @@ import models.supervision.Supervision
 import models.tcsp.Tcsp
 import models.tradingpremises.TradingPremises
 import models.tradingpremises.TradingPremises.FilterUtils
+import play.api.Logger
 import play.api.http.Status.UNPROCESSABLE_ENTITY
 import play.api.libs.json.{Format, Json}
 import uk.gov.hmrc.http.cache.client.CacheMap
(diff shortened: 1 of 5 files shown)


app/services/SubmissionService.scala
```diff
@@ -83,14 +83,28 @@ class SubmissionService @Inject()(val cacheConnector: DataCacheConnector,
...
       case e: Upstream4xxResponse if e.upstreamResponseCode == UNPROCESSABLE_ENTITY =>
         Future.failed(SubscriptionErrorResponse.from(e).fold[Throwable](e)(r => DuplicateSubscriptionException(r.message)))
     }
   }
 
+  private def getPostcodeFromTP(subscriptionRequest: SubscriptionRequest) = {
+
+    subscriptionRequest.tradingPremisesSection match {
+      case Some(tpSection) => {
+        tpSection.head.yourTradingPremises.map(x => x.tradingPremisesAddress) match {
```
slavvco, line 98: here instead of Some(tpSection) you could use 
Some(firstTp +: _) => 
or 
Some(Seq(firstTp, _ *)) =>
just to avoid using tpSection.head which can have similar effect to using get on option

ajchilts: @slavvco - I think the above is difficult to read. the user can't submit with zero trading premises so using head should be fine. If it raises an exception then good as we have an incomplete section.

PR #1495, opened 2020-11-03
DLS-3456: Added more logging for investigation.

rambabu-posa:


app/controllers/SubmissionController.scala
```diff
@@ -78,8 +88,15 @@ class SubmissionController @Inject()(val subscriptionService: SubmissionService,
           case _: BadRequestException =>
             Logger.info("[SubmissionController][post] handling BadRequestException")
             Future.successful(Ok(bad_request()))
+          case e: Exception =>
+            Logger.info("[SubmissionController][post] handling Exception")
+            throw new Exception(e.getMessage)
```
callum-gibbons, line 93: We probably just want to throw e here rather than creating a new generic exception

rambabu-posa: fixed it

PR #1516, opened 2022-07-18
Dls 5401

priyankadhiman3006:
[ticket details -> https://jira.tools.tax.service.gov.uk/browse/DLS-5401]
- Added new template v5m0.
- Added test cases for the change
- Addition of logic where heading and subject line for the new template can be differentiated from the old template

##As of now , the logic to identify which template to choose is not there . It will be covered in step 2  DLS_5402. So for testing data is inserted manually under amls-notification and then template data was tested

## Checklist

- [ X] Build passing.
- [X ] Unit tests have full coverage.
- [ ] RELEASE NOTES ADDED.


app/services/NotificationService.scala
```diff
@@ -116,19 +118,24 @@ class NotificationService @Inject()(val amlsNotificationConnector: AmlsNotificat
...
     amlsNotificationConnector.getMessageDetailsByAmlsRegNo(amlsRegNo, id, accountTypeId) map {
-      case Some(notificationDetails) => {
+      case Some(notificationDetails) =>
         for {
           message <- notificationDetails.messageText
           details <- NotificationDetails.convertEndDateMessageText(message)
         } yield {
+
+          val endDate: String =
+          if(templateVersion == "v5m0" && (contactType == ContactType.RenewalReminder || contactType == ContactType.NewRenewalReminder))
```
lutnos, line 128: Why does this have to be done especially for v5m0, but not for others, please?  Perhaps because content change has mandated it.  I would really prefer that it is dealt somewhere when the content is different for each version case.  

In any case, as above, it is not safe: a future v6 will revert to pre-v4 behaviour

app/models/notifications/NotificationDetails.scala
```diff
@@ -31,7 +32,14 @@ case class NotificationDetails(contactType: Option[ContactType],
 
   val cType = ContactTypeHelper.getContactType(status, contactType, variation)
 
-  def subject = s"notifications.subject.$cType"
+  def subject(templateVersion: String) = {
```
lutnos, line 35: This is inherently awkward and risky, what happens one day when v6 arrives, by default it will get the same messages as v1..4 .

I would be tempted to code this as something like:

templateVersion match {
case "v1m0" |  "v2m0" | "v3m0" | "v4m0" => s"notifications.subject.$cType
case "v5m0" => s"notifications.subject.new.$cType
case _ => throw new Exception s"Unknown template version $templateVersion"
}

Then have a notifications.subject.new.xxx for each and every cType, not just the renewalReminders.

The benefits are: that it will fail if future templateVersions are not addressed, the code here does not need to worry about which subjects change and which do not, the presumption will also be that they will not change each version after v5 (they did not so far)

lutnos (changes requested):
A couple of changes I would like you to consider, please

app/services/NotificationService.scala
priyankadhiman3006: This has too be done only for v5 because the date format in the inbox message has changed from YYYY-MM-DD to DD-MM-YYYY for renewal reminder and it shouldn't change for the old templates. This is  done here as this is the place we are passing end date in the template

priyankadhiman3006: Let me have a look if that can be done in MessageDetails.scala 👍

priyankadhiman3006: date change is now done at MessageDetails.scala for the two contactTYpe
https://github.com/hmrc/amls-frontend/commit/dbc622a525963869c2a520ceceab6155b8a8c2dc

app/models/notifications/NotificationDetails.scala
priyankadhiman3006: changes done 
https://github.com/hmrc/amls-frontend/commit/dbc622a525963869c2a520ceceab6155b8a8c2dc

lutnos: Thanks, you perhaps missed the subtlety that I suggested .new in place of .v5 because in all probability .v5 is going to equally apply to later versions too.  But that is no biggy, it is fine as you have done it

app/services/NotificationService.scala
```diff
@@ -52,8 +53,16 @@ class NotificationService @Inject()(val amlsNotificationConnector: AmlsNotificat
...
       case ContactType.RenewalApproval |
-           ContactType.AutoExpiryOfRegistration |
-           ContactType.RenewalReminder => handleEndDateMessage(amlsRegNo, id, contactType, templateVersion, accountTypeId)
+           ContactType.AutoExpiryOfRegistration => handleEndDateMessage(amlsRegNo, id, contactType, templateVersion, accountTypeId)
+
+      case ContactType.RenewalReminder |
+           ContactType.NewRenewalReminder => {
+        templateVersion match {
+          case "v5m0"  => handleEndDateMessage(amlsRegNo, id, contactType, templateVersion, accountTypeId)
+          case "v1m0" | "v2m0" | "v3m0" | "v4m0" | "v5m0 " => handleEndDateMessage(amlsRegNo, id, contactType, templateVersion, accountTypeId)
```
lutnos, line 62: This can't be right, you have included "v5m0" in this case too (I know it want ever match there, because it is lower, but anyway ...)

More importantly both matches return the same value:

handleEndDateMessage(amlsRegNo, id, contactType, templateVersion, accountTypeId)

What do you mean to do?

lutnos (changes requested):

app/services/NotificationService.scala
priyankadhiman3006: changes reverted here -> https://github.com/hmrc/amls-frontend/pull/1516/commits/115f2a10c48572c6bbbf6ac6ceb7070c7b5d509b

lutnos (approved):

PR #1520, opened 2022-08-04
fix heading

priyankadhiman3006:
fix heading issue

### DIFF
--- a/app/models/notifications/NotificationDetails.scala
+++ b/app/models/notifications/NotificationDetails.scala
@@ -18,6 +18,8 @@ package models.notifications
 
 import cats.implicits._
 import models.confirmation.Currency
+import models.notifications.ContactType._
+import models.notifications.NotificationDetails.isContactTypev5
 import org.joda.time.format.{DateTimeFormat, DateTimeFormatter, ISODateTimeFormat}
 import org.joda.time.{DateTime, DateTimeZone, LocalDate}
 import play.api.libs.json._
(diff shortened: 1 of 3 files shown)


app/models/notifications/NotificationDetails.scala
```diff
@@ -33,8 +35,8 @@ case class NotificationDetails(contactType: Option[ContactType],
 
   def subject(templateVersion: String) =
     templateVersion match {
-      case "v1m0" | "v2m0" | "v3m0" | "v4m0" => s"notifications.subject.$cType"
-      case "v5m0" => s"notifications.subject.v5.$cType"
+      case "v5m0" if(isContactTypev5(cType))=> s"notifications.subject.v5.$cType"
```
priyankadhiman3006, line 38: Instead of adding new message for all contact TYpe, I have added a logic that new heading applicable to only those contact types whose templates have changed otherwise use the old heading from message file.

lutnos (changes requested):
HI Priyanka, I really don't like this way of dealing with some subjects as a special case, I would much rather we dealt with this in a regular way - by having v5m0 versions of the message lines for each subject, please

priyankadhiman3006:
> HI Priyanka, I really don't like this way of dealing with some subjects as a special case, I would much rather we dealt with this in a regular way - by having v5m0 versions of the message lines for each subject, please

Changes done and all messages are displaying correctly

lutnos (approved):

PR #1522, opened 2022-11-11
DLS-5616 Mongo Library Upgrade

TomRafferty:
This PR navigates amls-frontend to use the new hmrc-mongo.

Unit and acceptance tests pass.

### DIFF
--- a/app/services/cache/MongoCacheClient.scala
+++ b/app/services/cache/MongoCacheClient.scala
@@ -19,20 +19,19 @@ package services.cache
 import config.ApplicationConfig
 import connectors.cache.Conversions
 
+import play.api.Logging
 import javax.inject.Inject
-import org.joda.time.{DateTime, DateTimeZone}
+import java.time.{LocalDateTime, ZoneOffset}
+import java.time.format.DateTimeFormatter
 import play.api.libs.json._
-import play.modules.reactivemongo.ReactiveMongoComponent
-import reactivemongo.api.DefaultDB
-import reactivemongo.api.commands.WriteResult
-import reactivemongo.api.indexes.{Index, IndexType}
-import reactivemongo.bson.{BSONDocument, BSONObjectID}
-import reactivemongo.play.json.ImplicitBSONHandlers._
 import uk.gov.hmrc.crypto.json.{JsonDecryptor, JsonEncryptor}
 import uk.gov.hmrc.crypto.{ApplicationCrypto, _}
 import uk.gov.hmrc.http.cache.client.CacheMap
-import uk.gov.hmrc.mongo.ReactiveRepository
-import uk.gov.hmrc.mongo.json.ReactiveMongoFormats
+import uk.gov.hmrc.mongo.MongoComponent
+import uk.gov.hmrc.mongo.play.json.PlayMongoRepository
+import uk.gov.hmrc.mongo.play.json.formats.MongoJavatimeFormats
+import org.mongodb.scala.model.{Filters, FindOneAndUpdateOptions, IndexModel, IndexOptions, Indexes, ReturnDocument}
+import org.mongodb.scala.bson.BsonDocument
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
@@ -41,7 +40,7 @@ import scala.util.{Failure, Success, Try}
 // $COVERAGE-OFF$
 // Coverage has been turned off for these types, as the only things we can really do with them
 // is mock out the mongo connection, which is bad craic. This has all been manually tested in the running application.
-case class Cache(id: String, data: Map[String, JsValue], lastUpdated: DateTime = DateTime.now(DateTimeZone.UTC)) {
+case class Cache(id: String, data: Map[String, JsValue], lastUpdated: LocalDateTime = LocalDateTime.now(ZoneOffset.UTC)) {
 
   /**
     * Upsert a value into the cache given its key.
@@ -57,13 +56,13 @@ case class Cache(id: String, data: Map[String, JsValue], lastUpdated: DateTime =
 
     this.copy(
       data = updated,
-      lastUpdated = DateTime.now(DateTimeZone.UTC)
+      lastUpdated = LocalDateTime.now(ZoneOffset.UTC)
     )
   }
 }
 
 object Cache {
-  implicit val dateFormat = ReactiveMongoFormats.dateTimeFormats
+  implicit val dateFormat = MongoJavatimeFormats.localDateTimeFormat
   implicit val format = Json.format[Cache]
 
   def apply(cacheMap: CacheMap): Cache = Cache(cacheMap.id, cacheMap.data)
@@ -86,7 +85,7 @@ class CryptoCache(cache: Cache, crypto: CompositeSymmetricCrypto) extends Cache(
 /**
   * An injectible factory for creating new MongoCacheClients
   */
-class MongoCacheClientFactory @Inject()(config: ApplicationConfig, applicationCrypto: ApplicationCrypto, component: ReactiveMongoComponent) {
+class MongoCacheClientFactory @Inject()(config: ApplicationConfig, applicationCrypto: ApplicationCrypto, mongo: MongoComponent) {
   def createClient: MongoCacheClient = new MongoCacheClient(config, component.mongoConnector.db, applicationCrypto)
 }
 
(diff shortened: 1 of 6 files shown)


app/services/cache/MongoCacheClient.scala
```diff
@@ -214,16 +242,23 @@ class MongoCacheClient(appConfig: ApplicationConfig, db: () => DefaultDB, applic
...
   /**
     * Removes the item with the specified id from the cache
     */
-  def removeById(credId: String) =
-    collection.delete().one(key(credId)) map handleWriteResult
+  def removeById(credId: String):Future[Boolean] = {
+    collection.findOneAndDelete(key(credId)).toFuture()
+      .map { result => true
+      }
+      .recover { case _ => false }
```
LAJW, line 258: ⛏️ Should we log this error or just not handle?

app/services/cache/MongoCacheClient.scala
```diff
@@ -214,16 +242,23 @@ class MongoCacheClient(appConfig: ApplicationConfig, db: () => DefaultDB, applic
...
+  }
 
   /**
     * Removes the item with the specified id from the cache
     */
-  def removeById(credId: String) =
-    collection.delete().one(key(credId)) map handleWriteResult
+  def removeById(credId: String):Future[Boolean] = {
+    collection.findOneAndDelete(key(credId)).toFuture()
+      .map { result => true
```
LAJW, line 256: ❓ Should we check for the number of deleted rows? Wouldn't this return true even if there were zero deletions?

app/services/cache/MongoCacheClient.scala
```diff
@@ -239,8 +274,14 @@ class MongoCacheClient(appConfig: ApplicationConfig, db: () => DefaultDB, applic
...
-
-    collection.update(ordered = false).one(bsonIdQuery(cache.id), BSONDocument("$set" -> Json.toJson(rebuiltCache)), upsert = true) map handleWriteResult
+    collection.findOneAndUpdate(
+      filter= bsonIdQuery(cache.id),
+      update = Updates.combine(
+        Updates.set("id", rebuiltCache.id),
+        Updates.set("data",Codecs.toBson(rebuiltCache.data)),
+        Updates.set("lastUpdated", LocalDateTime.now(ZoneOffset.UTC))),
+      options = FindOneAndUpdateOptions().upsert(true).returnDocument(ReturnDocument.AFTER)
+    ).toFuture.map(result => true)
```
LAJW, line 284: ❓ Should we replace this `true` with unit if it always succeeds?

app/services/cache/MongoCacheClient.scala
```diff
@@ -239,8 +274,14 @@ class MongoCacheClient(appConfig: ApplicationConfig, db: () => DefaultDB, applic
...
       }
     })
-
-    collection.update(ordered = false).one(bsonIdQuery(cache.id), BSONDocument("$set" -> Json.toJson(rebuiltCache)), upsert = true) map handleWriteResult
+    collection.findOneAndUpdate(
+      filter= bsonIdQuery(cache.id),
+      update = Updates.combine(
+        Updates.set("id", rebuiltCache.id),
+        Updates.set("data",Codecs.toBson(rebuiltCache.data)),
+        Updates.set("lastUpdated", LocalDateTime.now(ZoneOffset.UTC))),
```
LAJW, line 282: ❓ Why set fields individually, rather than do `Codecs.toBon(rebuildCache)`

app/services/cache/MongoCacheClient.scala
```diff
@@ -129,40 +148,45 @@ class MongoCacheClient(appConfig: ApplicationConfig, db: () => DefaultDB, applic
...
     fetchAll(Some(credId)) flatMap { maybeNewCache =>
       val cache = maybeNewCache.getOrElse(Cache(credId, Map.empty))
 
       val updatedCache = cache.copy(
         data = cache.data - (key),
-        lastUpdated = DateTime.now(DateTimeZone.UTC)
+        lastUpdated = LocalDateTime.now(ZoneOffset.UTC)
       )
 
       val document = Json.toJson(updatedCache)
```
LAJW, line 181: ⛏️ Unused variable.

app/services/cache/MongoCacheClient.scala
```diff
@@ -129,40 +148,45 @@ class MongoCacheClient(appConfig: ApplicationConfig, db: () => DefaultDB, applic
...
     */
   def removeByKey[T](credId: String, key: String): Future[Cache] = {
-
     fetchAll(Some(credId)) flatMap { maybeNewCache =>
       val cache = maybeNewCache.getOrElse(Cache(credId, Map.empty))
 
       val updatedCache = cache.copy(
         data = cache.data - (key),
-        lastUpdated = DateTime.now(DateTimeZone.UTC)
+        lastUpdated = LocalDateTime.now(ZoneOffset.UTC)
```
LAJW, line 178: ⛏️ This object is also unused.

app/services/cache/MongoCacheClient.scala
```diff
@@ -129,40 +148,45 @@ class MongoCacheClient(appConfig: ApplicationConfig, db: () => DefaultDB, applic
     }
 
     fetchAll(Some(credId)) flatMap { maybeNewCache =>
-
       val cache: Cache = maybeNewCache.getOrElse(Cache(credId, Map.empty))
 
+      val d= cache.data + (key -> jsonData)
```
LAJW, line 153: ⛏️ Unused.

LAJW (approved):
I made the mistake of reviewing commit by commit. Almost all of my comments would be eventually addressed 👍 
 - Ram is right about documenting that "24 hour" cache clear option 
 - This should be squashed
 - ⛏️ is a nitpick
 - ❓ is a question
Other than that, it looks good to me.

conf/application.conf
```diff
@@ -126,6 +128,7 @@ controllers {
 
 mongodb {
   uri = "mongodb://localhost:27017/amls-frontend"
+  mongocache.expireAfter = 24 hours
```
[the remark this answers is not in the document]
TomRafferty: looking into this...

app/services/cache/MongoCacheClient.scala
TomRafferty: As we are removing, the only fail case I guess is that the element we want to remove doesn't exist, in which case I guess that's the goal either way.

We could always log this result but I'm unsure if this is needed? 🤔

app/services/cache/MongoCacheClient.scala
TomRafferty: as mentioned in the previous comment, we could log that if it would be helpful but I think if the goal is to not have something in the db matching that id then either way the code has succeeded

app/services/cache/MongoCacheClient.scala
TomRafferty: will look into this

app/services/cache/MongoCacheClient.scala
LAJW: In that case, consider switching to returning `Unit` - it would tell you if there are any redundant checks when using this function ;)

app/services/cache/MongoCacheClient.scala
TomRafferty: after having a play with this idea, it's a bit tricky to get the lastUpdated param in there - there will be a way, however this is pretty straight forward :)

app/services/cache/MongoCacheClient.scala
TomRafferty: good spot, 
was left there from the old code 😬

app/services/cache/MongoCacheClient.scala
TomRafferty: looking into this as it's a little funky

app/services/cache/MongoCacheClient.scala
TomRafferty: good spot, more fragments of old code

app/services/cache/MongoCacheClient.scala
TomRafferty: Upon a closer look I believe this is used to make the use of cache.copy easier - this is to do with immutable types I believe.

LAJW: I meant: you can delete this `val updatedCache` here and just use `cache.data - key` (you're only using `updatedCache.data` below and not the whole `updatedCache`).

`.copy` creates a copy of an object with new properties assigned. Since you're only using one property below, which you're assigning here, you can skip this step.

```scala
val updatedData = cache.data - (key)
...
Codecs.toBson(updatedData)
// or
...
Codecs.toBson(cache.data - (key))
```

app/services/cache/MongoCacheClient.scala
LAJW: Logging in case there's something wrong with the connection or Mongo schema has changed (recover will fire if it can't parse the result for example).

app/services/cache/MongoCacheClient.scala
TomRafferty: This is a little more awkward as code within the controller depends on it being a bool - I think it may be to do with errors and such.

LAJW: Exactly. So my suggestion was to modify the controller code and only leave the "happy" path - assume this function returns true, delete - now dead - `else` branches.

app/services/cache/MongoCacheClient.scala
TomRafferty: Sorry got confused, I actually made that change and then for some reason said that it couldn't be done 🤦 
Maybe not enough caffeine in me... or too much? 🤔

PR #1586, opened 2024-01-23
AMLS page guidance added

ellie-l-newell:

## Checklist

- [ ] Build passing.
- [ ] RELEASE NOTES ADDED.


app/views/businessmatching/PsrNumberView.scala.html
```diff
@@ -121,16 +130,20 @@
 
                 @subtitle(section)
 
+                @heading("businessmatching.psr.number.title")
```
jpatton638, line 133: Might be the case that this content needs to be added for the above case statement too, have you checked what the difference is between this case and the one above?

jpatton638 (approved):

PR #1613, opened 2024-04-19
DDCYLS-3042 Update various depedencies, removal of some unused

lewishay:
I didn't touch bootstrap or play-partials as these both bring in a later version of crypto which causes problems addressed in another ticket (no idea why play-partials does this).

play-frontend-hmrc was upgraded with the help of the guidance in the library's changelog. One notable change is that it has changed the capitalisation of the content inside govuk tags, so for AMLS this is the row statuses such as Completed, Updated, etc. Because of this there is also an AT PR as it is matching on the tag content: https://github.com/hmrc/amls-acceptance-test/pull/368

Most of the test changes are due to removal of explicit mockito dep in favour of using the one provided by bootstrap. I also addressed a few warnings in the service but didn't go too far with it to prevent overwhelming the PR.

### DIFF
--- a/project/AppDependencies.scala
+++ b/project/AppDependencies.scala
@@ -1,4 +1,3 @@
-import play.core.PlayVersion
 import play.sbt.PlayImport.ws
 import sbt.*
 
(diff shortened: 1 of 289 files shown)


project/AppDependencies.scala
```diff
@@ -42,12 +38,7 @@ private object AppDependencies {
         "uk.gov.hmrc"            %% s"bootstrap-test-$playV"  % bootstrapV          % scope,
         "uk.gov.hmrc.mongo"      %% s"hmrc-mongo-test-$playV" % hmrcMongoV          % scope,
         "org.scalatestplus.play" %% "scalatestplus-play"      % "7.0.1"             % scope,
```
jpatton638, line 40: This can be removed, this version is pulled in by bootstrap

jpatton638 (approved):

PR #1621, opened 2024-04-30
DDCYLS-4581

dips7189:
== AppDependencies ==
- Removed json-encryption library
- Added crypto-json library

== build.sbt ==
- Upgraded Bootstrap Version from 6.0.0 to 7.3.0

== ApplicationConfig ==
- Added function to retrieve the JSON Encryption Key

== Crypto ==
- New file to contain the wrapper that holds decrypted values
- This is required by crypto-json

== Country ==
- Added United Kingdom value to be used anywhere rather than
writing whole class

== Cache ==
- This previously lived in MongoCacheClient, it has been moved out
into it's own file

- tryDecrypt function has been added to do the same thing as getEntry
function but it returns a Try to hold the thrown exception rather than
just throwing it for better code flow

- sanitiseDoubleDecrypt function has been added to fulfill two purposes:
The first is to perform encryption twice on values that have been
accidentally encrypted twice.
The second is to sanitise incoming input which maybe a JSON array character
or JSON Object charatcer which for some reason are stored as literal strings

== CacheOps ==
- CompositeSymmetricCrypto has been deprecated in the new crypto-json library,
Encrypter with Decrypter is used instead.
- decryptValue now uses the crypto-json library specific features.

== CryptoCache ==
- This file previously lived in MongoCacheClient, but has since been moved out
to its own external file.

== MongoCacheClient ==
- tryDecrypt has been deleted. This function was previolusly decrypting to JSON Strings,
however since the upgrade it was failing to do so due to the library internals moving
away from Apache Base64 decoding to regular Java Base64 decoding, as such decryption was
not happening & the encrypted value was being returned & saved in higher calls to this
function.

- CompositeSymmetricCrypto has been deprecated in the new crypto-json library & has been replaced
with Encrypter with Decrypter

- EncryptionService is now responsible for taking care of all JSON encryption & decryption
rather than it happening in this class.

== EncryptionService ==
- Created to perform the following:

1. Encryption
2. Decryption
3. Base64 Encoding
4. Base64 Decoding

The real problem migrating to this library was that the previously used Apache Commons library
that used a Base64 encoding scheme that would enter newlines & carriage returns to the end of the
encrypted then encoded value. The new crypto-json library uses Java to decode Base64 values
& rejects anything outside of the Base64 alphabet as stated in the RFC-4648 documentation. Those
carriage returns & newline characters are rejected & an exception kept being thrown.

This service uses the Apache Commons library to carry with the encoding & decoding rather than using
the crypto-json library's way of doing it.

This service is contains an underlying implementation that is not strictly tied to any encryption/decryption
library.

== ControllerHelper ==
- hasInvalidRedressScheme has been removed since it does the same thing as the exist function
available in the option data type.

== CacheValues ==
- New file to hold test data

== LandingControllerWithAmendmentSpec ==
- massive refactor of tests

== CacheSpec ==
- New test class added

== EncryptionServiceSpec ==
- New test class added

### DIFF
--- a/test/controllers/LandingControllerWithAmendmentsSpec.scala
+++ b/test/controllers/LandingControllerWithAmendmentsSpec.scala
@@ -284,8 +284,6 @@ class LandingControllerWithAmendmentsSpec extends AmlsSpec with MockitoSugar wit
                 when(controller.statusService.getDetailedStatus(any(), any[(String, String)], any())(any[HeaderCarrier](), any(), any()))
                   .thenReturn(Future.successful((NotCompleted, None)))
 
-                val updatedTestCacheMap = testCache.data + (ResponsiblePerson.key -> Json.toJson(Seq.empty[ResponsiblePerson]))
-
                 val result = controller.get()(requestWithHeaders("test-context" -> "ESCS"))
 
                 status(result) must be(SEE_OTHER)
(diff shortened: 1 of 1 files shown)


lewishay (changes requested):

lewishay (approved):
Currently approved, waiting until the most recent change has been deployed before I merge

PR #1625, opened 2024-05-07
Removal of hidden label and duplication of title. Amended hint text

ellie-l-newell:

## Checklist

- [ ] Build passing.
- [ ] RELEASE NOTES ADDED.


app/views/businessactivities/BusinessEmployeesAMLSSupervisionView.scala.html
```diff
@@ -56,9 +54,7 @@
                 inputmode = Some("numeric"),
                 classes = "govuk-!-width-one-quarter",
                 label = Label(
-                    isPageHeading = false,
-                    content = Text(messages("businessactivities.employees.amls.supervision.title")),
-                    classes = "govuk-visually-hidden"
+                    isPageHeading = false
```
lewishay, line 57: This doesn't quite achieve the markup we want. Instead of having an empty label the idea would be to have the h1 as the label, and inside of the form group - so move the heading content into here and set this boolean to true. You may also need a specific class here too to make it look like a heading. Same applies for the other page in this PR. If unsure inspect the HTML in the browser when loading the page locally and compare with the HTML a similar example on the design system: https://design-system.service.gov.uk/components/text-input/

lewishay (changes requested):

app/views/businessactivities/BusinessEmployeesAMLSSupervisionView.scala.html
ellie-l-newell: Made the changes for this and other page in PR

lewishay (approved):
Approved, awaiting clear pipeline before merge

PR #1626, opened 2024-05-07
DDCYLS-5047 adding logging in double decryption functions

dips7189:


app/services/cache/CacheOps.scala
```diff
@@ -49,6 +50,7 @@ trait CacheOps {
   def catchDoubleEncryption[T](cache: Cache, key: String)(implicit reads: Reads[T], c: Encrypter with Decrypter): Option[T] = {
     Try(decryptValue[T](cache, key)(reads, c)) match {
       case Failure(_: JsResultException) =>
+        logger.info(s"performing double decryption")
```
lewishay, line 53: As mentioned on Slack our root logger level in Production is set to `warn` so the added statements need to be warn or higher

lewishay (changes requested):

lewishay (approved):

PR #1681, opened 2024-08-01
DDCYLS-4657 - Validation of passport number does not allow spaces

Liam-Macarthur:
Trim and removal of white space in passport number validation.

### DIFF
--- a/app/forms/responsiblepeople/PersonNonUKPassportFormProvider.scala
+++ b/app/forms/responsiblepeople/PersonNonUKPassportFormProvider.scala
@@ -36,7 +36,9 @@ class PersonNonUKPassportFormProvider @Inject()() extends Mappings {
       booleanFieldName -> boolean(emptyError, emptyError),
       "nonUKPassportNumber" -> mandatoryIfTrue(
         booleanFieldName,
-        text(s"$emptyError.number").verifying(
+        text(s"$emptyError.number")
+          .transform[String](_.replace(" ", "").trim, x => x)
+          .verifying(
           firstError(
             maxLength(length, "error.invalid.non.uk.passport.number.length.40"),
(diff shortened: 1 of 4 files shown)


app/forms/responsiblepeople/PersonNonUKPassportFormProvider.scala
```diff
@@ -36,7 +36,9 @@ class PersonNonUKPassportFormProvider @Inject()() extends Mappings {
       booleanFieldName -> boolean(emptyError, emptyError),
       "nonUKPassportNumber" -> mandatoryIfTrue(
         booleanFieldName,
-        text(s"$emptyError.number").verifying(
+        text(s"$emptyError.number")
+          .transform[String](_.replace(" ", "").trim, x => x)
```
jpatton638, line 40: There is a function that has just gone in to main that would be very handy for this:

https://github.com/hmrc/amls-frontend/blob/main/app/forms/mappings/Mappings.scala#L53

If you could rebase and use this function that would be great!

app/forms/responsiblepeople/PersonUKPassportFormProvider.scala
```diff
@@ -38,13 +38,15 @@ class PersonUKPassportFormProvider @Inject()() extends Mappings {
       booleanFieldName -> boolean(booleanError, booleanError),
       "ukPassportNumber" -> mandatoryIfTrue(
         booleanFieldName,
-        text(s"$booleanError.number").verifying(
-          firstError(
-            maxLength(length, lengthError),
-            minLength(length, lengthError),
-            regexp(regex, "error.invalid.uk.passport")
+        text(s"$booleanError.number")
+          .transform[String](_.replace(" ", "").trim, x => x)
```
jpatton638, line 42: As above

PR #1688, opened 2024-08-06
DDCYLS-5562 - differentiate between registration and renewal journey for bacs confirmation screen

lewis-nicholls:
As part of the AMLS registration and renewal journeys, the shared bacs confirmation screen now differentiates between the journey, displaying a new panel heading and text for the link to return to the dashboard.


VJWatson (approved):
only thing i can see is adding a space after the word ok on line 47 , but i don't believe it would break anything

dnyuar (approved):
Approved

PR #1787, opened 2026-07-15
DDCYLS-8730  AMLS Euros to Pounds PR016402-REQ021942 - UPDATE NOTIFICATION BANNER STYLING

Antony100:


VJWatson (approved):

app/views/renewal/PercentageView.scala.html
```diff
@@ -47,7 +47,7 @@
 
     @errorSummary(form, Some(Map("percentage" -> PercentageOfCashPaymentOver15000.all.head.toString)))
```
VJWatson, line 48: are these ones on all the time? Dont they all need to be in an If?

Antony100: Yes, if I have understood correctly. These are part of the renewal journey, in the Mural the instruction says: The banners should only appear on the renewal & change versions of the pages, not in the application journey

